PR #16, opened 2017-09-24
Poloniex: Fix Base/Counter inversion to subscribe channel

pleandre:
Base and counter currency were inverted because of a wrong conversion during channel subscription, I changed to use the conversion from PoloniexUtils.toPairString(currencyPair); which is the right one


pleandre:
In fact, in poloniex contrary to most exchanges the format is  COUNTERCCY_BASECCY (e.g. In poloniex utils, to pair string is correct) :
https://github.com/timmolter/XChange/blob/master/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexUtils.java

PR #40, opened 2017-12-06
Add isAlive method to StreamingExchange and underlying services

roeiba:
- Add isSocketOpen to
  - Add isAlive at StreamingExchange interface
  - Implement isAlive at poloniex2, wex, bitstamp and bitfinex

It allows user to check whether websocket is opened and handle disconnections. It also makes the 'disconnect' command safe by not writing to an already disconnected channel.

Using this API the user can check connection status and reconnect if needed, then user can resubscribe to all channels.

Sometimes subscribing to channel could raise exception and kill the connection (e.g channel not exist on exchange), so I prefered not to create re-subscription mechanism at service level, but leave it to the user. 

This is on top of the disconnect event, as it allows user to check connection status in sync manner as well.

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -129,11 +129,13 @@ public abstract class NettyStreamingService<T> {
 
     public Completable disconnect() {
         return Completable.create(completable -> {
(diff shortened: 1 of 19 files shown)


roeiba:
Code was tested and verified on all implemented exchanges.

dozd:
Rebased and fixed build error, but there are some exchanges left with // TODO Auto-generated method stub

roeiba:
Update code.

PR #51, opened 2017-12-18
Binance streaming order book

hkscfreak:
Example is from API docs:
https://www.binance.com/restapipub.html#depth-wss-endpoint


dozd:
Can you update README as well and mark support for this exchange? It would be super cool to have it in this PR :). THanks

hkscfreak:
Done!

PR #68, opened 2018-01-06
Add Binance isAlive implementation

roeiba:
This is done using same logic of Gemini who also uses ProductStreamingService

### DIFF
--- a/xchange-stream-core/pom.xml
+++ b/xchange-stream-core/pom.xml
@@ -15,6 +15,12 @@
             <artifactId>service-core</artifactId>
             <version>${project.parent.version}</version>
         </dependency>
+        
+        <dependency>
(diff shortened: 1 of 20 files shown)


xchange-stream-core/pom.xml
```diff
@@ -15,6 +15,12 @@
             <artifactId>service-core</artifactId>
             <version>${project.parent.version}</version>
         </dependency>
+        
+        <dependency>
```
dozd: this should not be in core as dependency. These services are only for exchange implementations. Why it is here?

roeiba: POM file was pushed by mistake. Please ignore it.

dozd:
Pls rebase onto develop

dozd:
No response from the user. Closing as not merged.

PR #77, opened 2018-01-16
[Binance] Adds ticker support

janakerman:
This adds ticker support to the BinanceExchange.

Previously, we were opening a separate connection for every product we subscribe to. Now we open and maintain a single connection. Due to Binance's API we have to subscribe to all of these up front via query params.

Their socket connection begins sending the pairs instantly, so we need to register subscribers to the observables to force the Observable body to register the messages for the incoming channels, to avoid NPE.

We cache these shared subscriptions, and return them when subscriptions on the MarketDataService are requested.

I wanted to put this up as a work in progress. This approach has the pros of having a single WS connection, but the cons of feeling a little shoe-horned into JSONNettyStreamingService - as that is really geared for bidirectional messaging.

TODO:
- Adds tests
- Chase down merge request to XChange - as I have currently C+P a class while I wait for publicity of a constructor to be updated. We will also need to wait for the next release of XChange too?
- Tidy up a little


dozd:
Rebased onto develop, still failed test. Could you please look into it?

dozd:
And yes, I would like to have dependency only on release Xchange versions to maintain better stability and didn't need to wait with release for Xchange

janakerman:
Completely understand. My pull request has been merged to XChange so when that's done I'll tidy this up and finish my TODOs. Put it up for any early feedback on the approach. Thanks for taking a look!

dozd:
I am ok with PR and happy that `ProductSubscription` is useful in another exchange beside gdax

janakerman:
Great - waiting on [XChange PR](https://github.com/timmolter/XChange/pull/2174)!

janakerman:
@dozd - Fixed the tests but I'm still waiting on a release from xchange before this can really be merged.

janakerman:
@dozd - xchange recently released the required fixes. I've rebased and tests pass, example also runs as expected.

janakerman:
Timestamps aren't currently populated as the model object being used in the `Binance24h` from XChange. This has no concept of a timestamp. I'll fix this as an additional PR if possible

PR #78, opened 2018-01-17
Add handler for compressed websocket messages

kaiserfr:
This handler was necessary because Bitmex testnet suddenly startet to send compressed messages.


dozd:
This change was reverted by #89 . It must be added with configuration style to prevent breaking of other exchanges.

PR #89, opened 2018-02-02
Repair reconnecting

nodarret:
- repair reconnecting 
- logging with exchange name instead NettyStreamingService
- removed WebSocketClientCompressionHandler.INSTANCE handler


GlebYarnykh:
Should we consider blockingAwait logic like here https://github.com/firegnome/xchange-stream/commit/5d09ad5e2d299bcdaf2c91e1d786cc59ec1540e3?

I mean if we insert .blockingAwait after connect. in line 305?

nodarret:
I think timeout execption should be from netty, not completable (possible memory leaks?). Also every blocking mechanism in async programing is not recommended - both netty and rxJava are async. Better way is register action on event. In my case retryWhen retry with delay when Exception is thrown and doOnComplete resubscribes if completable completes.

GlebYarnykh:
So that basically means that your pull request make a valid reconnect only if exchange makes a nofitfication about websocket closing?

In case of GDAX, i do not recieve anything about websocket closing, just it silently drop and stopped recieving any messages..

Issue #80, opened 2018-01-19
Netty Websocket stability discussion


firegnome:
A month ago I committed an automatic reconnect for the poloniex websocket https://github.com/bitrich-info/xchange-stream/pull/43. This seems to work and I'm glad to see it integrated in the Netty class https://github.com/bitrich-info/xchange-stream/pull/56. This reconnect solves a lot of problems and is in my opinion the correct behaviour for the subscription.

Now we're still facing some stability problems with the websocket that aren't solved by the fix above.
1. The exchange drops the subscription on server side but doesn't inform us. (websocket stays open but no trades are sent)
2. A connection interrupt doesn't close the Netty websocket and doesn't trigger the disconnected event. 

My first idea was to enable tcp keepalive on the netty socket to recognize a connection lost, but this doesn't seem to solve the problem as described here: https://stackoverflow.com/questions/23238319/websockets-ping-pong-why-not-tcp-keepalive.

What I've learned is that we should use a custom ping pong but since poloniex doesn't provide such a functionality we have to use their own implementation. They solved this problem with heartbeats. @chuggol has also recognized this problem on gdax https://github.com/bitrich-info/xchange-stream/issues/66. 

Poloniex API heartbeats:
> In some markets, if there is no update for more than 1 second, a heartbeat message consisting of an empty argument list and the latest sequence number will be sent. These will go out once per second, but if there is no update for more than 60 seconds, the heartbeat interval will be reduced to 8 seconds until the next update.

What I ended up doing is to start a watcher thread that verifies that in the last 10 seconds at least a heartbeat or a trade/orderbook/... update was sent to verify that the websocket is still open. https://github.com/bitrich-info/xchange-stream/compare/develop...firegnome:develop#diff-0a517b95d1307fce62ef049280dd36c2R166

When the thread recognizes that the connection is lost it does following:
1. call the **resubscribe** function -> that will result in a connection timeout (I set the Netty websocket timeout to 10 seconds: https://github.com/bitrich-info/xchange-stream/compare/develop...firegnome:develop#diff-24429d40711776b7bfbbf1d0a497fd6eR118)
2. With this ReadTimeout Netty will recognize the lost connection and trigger the disconnected event
3. Now the reconnect/resubscribe procedure that was implemented in https://github.com/bitrich-info/xchange-stream/pull/43 will reestablish the connection.

[...]

PR #94, opened 2018-02-06
Generic WebsocketHealthWatcher thread

firegnome:
Implementation for https://github.com/bitrich-info/xchange-stream/issues/80

Since there is no generic way to find out if the websocket is still healthy, i wrote a method (isWebsocketReconnectRequired) that indicates if the socket needs to be reconnected. I implemented this for poloniex2 and its working for me.

@nodarret I tried my best to integrate your changes from #89 but I would really appreciate it If you could look at my changes.

As @hlbkin mentioned, there is a problem with the orderbooks:

> The only problem I see is that when you reconnect, orderBooks become stale i.e. need to be reconnected and submitted full cycle (GDAX full channel for example or binance).

This problem isn't solved yet but i propose to create a new ticket, since this isn't really caused by this pull request.


hlbkin:
For Bitfinex, they have explicit messages that state when you need to disconnect (https://bitfinex.readme.io/v2/docs/ws-general). is your pull request == workaround for that? or you have seen disconnects from Bitfinex WITHOUT this explicit message?

firegnome:
@hlbkin There are disconnects WITHOUT this explicit message, for example if you connect to a VPN Network it will drop the connection without message.

dozd:
Hello @firegnome , thank you for the PR. I'll go through it after my return from the vacation. It seems a little bit complicated on the first look so I'd like to understand it more before merging.

The main question is about reusability of the code for other exchanges. Is this the common pattern that will be useful elsewhere?

Second concern is about using `new Thread()` with `Thread.Sleep` in "rx world". Some sort of rx timer or interval would be preferred to keep code consistent.

But both are not crucial.

firegnome:
Hi @dozd 
To your first question: It is designed to be reusable for other exchanges. We had a discussion in #80  and other people confirmed, that this is needed on other exchanges too. 

Thank you for your hint with rx timer i will have a look at this and update the PR.

nodarret:
When I have free time I wil check code. For now im trying speed up makret data #101 or write my own solutnion ;)

hleb-albau:
General case -> start any stream, turn off Internet for 10min, turn on Internet. Application will be zoombied.

Flemingjp:
Is a heartbeat a given endpoint? It seems like an over engineered ping/pong transaction - which has now been implemented properly

badgerwithagun:
Well, the ping/pong now implemented is more so that the exchanges don't disconnect from their end. It doesn't tell us if a connection is up but has "gone quiet" - we're not sending pings, just responding to them.

I've never seen it where a socket is open but the exchange has stopped sending data. I HAVE seen it where we have reconnected but failed to send resubscription messages and the connection just starts cycling up and down.

I suspect this predates the relatively reliable auto reconnection logic, but it's hard to say.

badgerwithagun:
Heads up: I'm going to close this PR this week unless anyone feels strongly otherwise.

PR #95, opened 2018-02-07
Making streaming ThreadFactory aware

Devski:
Hi guys,

I'd like to enable streaming exchange users to set custom thread factory.

The change:
Implementing `setThreadFactory(ThreadFactory threadFactory)` method for `StreamingService`s that have `NettyStreamingService`.
The method throws `NotYetImplementedForExchangeException` for `StreamingService`s that do not `NettyStreamingService`.


dozd:
Hello, thank you for your contribution to the project.
I don't require any specific code style (as I make format commits before releasing) for the project but commits should contain only the necessary changes that are needed to implement / fix something.
Your commit is full of white-space changes - could you fix that please? It's unacceptable in current state - hard to review as well as bad for keeping proper track of history why lines of code were changed. Thanks!

PR #96, opened 2018-02-09
Overwrite NettyStreamingService.connect() in BitmexStreamingService

kaiserfr:
- This overwrite is necessary in order to add a WebSocketClientCompressionHandler. 
  This handler is required to handle compressed messages coming back from Bitmex websocket.

- In order to allow overwriting the connect() method the visibility of some members in NettyStreamingService has to be changed from private to protected


boppleton:
this has broken bitmex on my end, causing it to throw this on connection:

`io.netty.handler.codec.CodecException: invalid WebSocket Extension handshake for "permessage-deflate; client_no_context_takeover"`

is there a bitmex-specific config that i need to use to be compatible with this?  commenting out the add handler line (or the entire override method) makes it return to normal.

thebubbleindex:
I was able to at least get rid of this error by changing to the lines below in the info.bitrich.xchangestream.service.netty.NettyStreamingService class.

`WebSocketClientExtensionHandler clientExtensionHandler = null;`
`List<ChannelHandler> handlers = new ArrayList<>(4);`
`handlers.add(new HttpClientCodec());`
`if (false) handlers.add(WebSocketClientCompressionHandler.INSTANCE);`
`handlers.add(new HttpObjectAggregator(8192));`

`if (clientExtensionHandler != null) {`
`  handlers.add(clientExtensionHandler);`
`}`

The are two places in this code piece where WebSocketClientCompressionHandler.INSTANCE is being added as handler. It was originally: `WebSocketClientExtensionHandler clientExtensionHandler = getWebSocketClientExtensionHandler();` where this method returned the `INSTANCE` enum. Also, I turned off this line: `handlers.add(WebSocketClientCompressionHandler.INSTANCE);` with the `if (false)`

After doing this I was able to get JSON replies from Bitmex. However, it now seems there is an additional issue where Bitmex expects XBTUSD rather than BTCUSD.

PR #99, opened 2018-02-12
Update README to reflect availability of Binance trade/ticker web socket APIs

badgerwithagun:


janakerman:
Binance's API may support it, but it isn't yet implemented in this library yet. A PR is in for tickers - see #77.

Issue #120, opened 2018-03-05
[Binance] implement streaming trades


daniel-cohen:
API reference: [https://github.com/binance-exchange/binance-official-api-docs/blob/master/web-socket-streams.md#aggregate-trade-streams](https://github.com/binance-exchange/binance-official-api-docs/blob/master/web-socket-streams.md#aggregate-trade-streams)

I think we can get use the aggregated steam as we're not tracking individual trades anyway.

aleks-azen:
I think individual trades would be useful as well, if you want to track order completion without polling the api

daniel-cohen:
@aleks-azen OK, I see what your use-case is now.
Would you be open to maybe add a switch to BinanceStreamingExchange.connect to indicate whether you want the aggregated trades or single trades to come through:
```Observable<Trade> getTrades(CurrencyPair currencyPair, Object... args);```

I'm just trying to think how we can expose both the aggTrades and single trades..

**Comment by aleks-azen** [comment]
My primary experience is with the gdax and binance apis. The gdax api streams trades on an order by order basis. I assume that is likely the case for most exchanges but correct me if I'm wrong. To remain consistent I would suggest adding an aditional field to the StreamingMarketDataService interface with something like the following

```
   default Observable<AggTrade> getAggregatedTrades(CurrencyPair var1, Object... var2){
        throw new NotYetImplementedForExchangeException();
```

Then just building based around that with AggTrades and Trades being treated as entirely separate object. Using default would save you the hassle of implementing the function for every exchange.

**Comment by aleks-azen** [comment]
My primary experience is with the gdax and binance apis. The gdax api streams trades on an order by order basis. I assume that is likely the case for most exchanges but correct me if I'm wrong. To remain consistent I would suggest adding an aditional field to the StreamingMarketDataService interface with something like the following
```
   default Observable<Trade> getAggregatedTrades(CurrencyPair var1, Object... var2){
        throw new NotYetImplementedForExchangeException();
```
Also just add a new list of subscription types to the product subscription class. If an AggTrade is different enough from a trade for your needs it might be worth it to make a seperate AggTrade class with different fields

daniel-cohen:
@aleks-azen 
I think you're right and we'll need to implement both. Bitstamp seems to send individual trades too.
Bitfinex doesn't send any order ids (or any ids), so it could be aggregated.) see:
https://bitfinex.readme.io/v2/reference#ws-public-trades 

[...]

PR #121, opened 2018-03-05
Binance trades implementation

daniel-cohen:
Issue for tracking:
https://github.com/bitrich-info/xchange-stream/issues/120

### DIFF
--- a/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
+++ b/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
@@ -67,7 +67,7 @@ public class BinanceStreamingExchange extends BinanceExchange implements Streami
     private static String buildSubscriptionStreams(ProductSubscription subscription) {
         return buildSubscriptionStrings(subscription.getTicker(), "ticker") +
                 buildSubscriptionStrings(subscription.getOrderBook(), "depth") +
-                buildSubscriptionStrings(subscription.getTrades(), "trade");
+                buildSubscriptionStrings(subscription.getTrades(), "aggTrade");
     }
 
     private static String buildSubscriptionStrings(List<CurrencyPair> currencyPairs, String subscriptionType) {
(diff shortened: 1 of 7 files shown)


xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
```diff
@@ -67,7 +67,7 @@ private BinanceStreamingService createStreamingService(ProductSubscription subsc
     private static String buildSubscriptionStreams(ProductSubscription subscription) {
         return buildSubscriptionStrings(subscription.getTicker(), "ticker") +
                 buildSubscriptionStrings(subscription.getOrderBook(), "depth") +
-                buildSubscriptionStrings(subscription.getTrades(), "trade");
+                buildSubscriptionStrings(subscription.getTrades(), "aggTrade");
```
aleks-azen: the standard trade stream may still be implemented, there should be a new field for aggTrades

xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -61,6 +67,14 @@ public BinanceStreamingMarketDataService(BinanceStreamingService service) {
         }
         return tickerSubscriptions.get(currencyPair);
     }
+    
+    public Observable<BinanceAggTrades> getRawTrades(CurrencyPair currencyPair, Object... args) {
```
aleks-azen: same as above this should be getRawAggTrades

xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -70,7 +84,8 @@ public BinanceStreamingMarketDataService(BinanceStreamingService service) {
 
     @Override
     public Observable<Trade> getTrades(CurrencyPair currencyPair, Object... args) {
-        throw new NotYetImplementedForExchangeException();
+      return getRawTrades(currencyPair)
+            .map(at -> new Trade(BinanceAdapters.convertType(at.buyerMaker), at.quantity, currencyPair, at.price, at.getTimestamp(), Long.toString(at.aggregateTradeId)));
```
aleks-azen: same as above this will cause confusion if the standard trade stream is implemented

daniel-cohen: I have no problem implementing the "getRawAggTrades" as you suggested above. But I still think the override of:
```Observable<Trade> getTrades(CurrencyPair currencyPair, Object... args);```
Should return the aggregated trades here (Since we have to pick on if we implement both).

Maybe I'm missing something:
If we're not exposing (individual) trade ids , what's the point emitting single trades here ?

badgerwithagun:
@Flemingjp , I'm pretty sure this can just be closed...

PR #130, opened 2018-03-13
[xstream] Bumping dependency from 4.3.3 to 4.3.4.

bryantharris:
It would be nice to get some concrete releases that track with the underlying Xchange product.  From what I can see the latest streaming release is 4.3.1 where as XChange is up to 4.3.4.

Could a 4.3.4 release get pushed and have dev be placed on 4.3.5-SNAPSHOT?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
     </repositories>
 
     <properties>
-        <xchange.version>4.3.3</xchange.version>
+        <xchange.version>4.3.4</xchange.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <maven.compiler.source>1.8</maven.compiler.source>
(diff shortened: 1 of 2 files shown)


caleb-allen:
There are definitely upstream changes, like the potential addition of `getTickers` timmolter/XChange/issues/2385, that I would like to implement on xchange-stream but don't want to stray from the way the underlying project might implement it. Would be nice to know what general strategy xchange-stream takes to maintain parity.

dozd:
I'd like to have stable version of Xchange in develop to have stable build and release accordingly. Xchange-stream should follow latest major / minor version, but not patch (*.*.XXX). If you need SNAPSHOT dep on xchange, I would suggest to place it in different branch (name like upcomming, snapshot, dunno?) and merge it into develop once Xchange is released.
I will make new release of Xchange-stream because it's some time and yeah we are failing by a few patch version behind xchange

bryantharris:
As a suggestion, have the version of -stream be the same as the version xchange.  It just seems like an unnecessarily confusing approach to have two different version numbering systems for something so related.

dozd:
Solved in #156

Issue #81, opened 2018-01-20
CodecException: invalid WebSocket Extension handshake for "permessage-deflate; client_no_context_takeover"


gurumeditator:
Hi, I am attempting to use the 4.3.2-SNAPSHOT of the GDAX Streaming library. Upon connection to the websocket, I receive the error:

```
09:57:09.137 [main] INFO  info.bitrich.xchangestream.gdax.GDAXStreamingService - Registering GDAXWebSocketClientHandler
io.netty.handler.codec.CodecException: invalid WebSocket Extension handshake for "permessage-deflate; client_no_context_takeover"
	at io.netty.handler.codec.http.websocketx.extensions.WebSocketClientExtensionHandler.channelRead(WebSocketClientExtensionHandler.java:111)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.fireChannelRead(CombinedChannelDuplexHandler.java:438)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:310)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:284)
	at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:253)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1267)
	at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1078)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:489)
[...]

PR #139, opened 2018-03-19
add option for compression handler

boppleton:
some exchange websockets (ie. Bitmex Testnet) return compressed messages.  this has caused some exchanges to break from bad handshakes caused by the extension handlers for permessage-deflate.

this pr will allow you to optionally call exchange.useCompressedMessages(true) before calling connect() if you need to have the extension handlers added in your specific case

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -331,7 +331,5 @@ public abstract class NettyStreamingService<T> {
         return webSocketChannel.isOpen();
     }
 
-    public void useCompressedMessages(boolean compressedMessages) {
-        this.compressedMessages = compressedMessages;
-    }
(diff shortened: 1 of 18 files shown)


boppleton:
addresses #81 and reverts #96

TobCar:
@dozd Is there a way to load this pull request via a snapshot?

I'm trying to use

`<dependency>
    	<groupId>info.bitrich.xchange-stream</groupId>
    	<artifactId>xchange-stream-core</artifactId>
    	<version>4.3.3-SNAPSHOT</version>
	</dependency>
	<dependency>
    	<groupId>info.bitrich.xchange-stream</groupId>
    	<artifactId>xchange-bitmex</artifactId>
    	<version>4.3.3-SNAPSHOT</version>
</dependency>`

with `<repository>
  		<id>sonatype-oss-snapshot</id>
  		<snapshots/>
  	<url>https://oss.sonatype.org/content/repositories/snapshots</url>
	</repository>`

but it's not loading the content from this PR.

davidjirovec:
Build failed https://travis-ci.org/bitrich-info/xchange-stream/builds/360188701

PR #143, opened 2018-03-21
Feature/hitbtc

pchertalev:
Hi guys, let me provide implementation for HitBTC exchange (https://hitbtc.com/exchange/)


dozd:
Great work, thanks! You can update readme as well :)

PR #147, opened 2018-03-28
[Cexio] Add new exchange with private streaming

SergeyLukashevich:
All implemented streams are Cexio specific and uses its private API.
Main goal of changes is orders execution monitoring.

* Add Cexio authorization
* Add limit orders status update stream
* Add transactions stream


dozd:
Hello and thanks for the contribution. I will merge this but have some questions. This is the first implementation of private API in this project and it could be also a new building block for other implementations as well.

Tell me if you are willing to make a little more. I think of some interface for private data (like `StreamingPrivateDataService`) that could be useful to follow also on another exchanges. It could be accessible next to the `getStreamingMarketDataService()`. So the usage of private data part would be more straight-forward. What do you think?

SergeyLukashevich:
I'm agree that generic interface for private API is needed. Most likely later it have to be rewritten, since some exchanges offer not only read-only data streaming, but trading and other operations, but I propose `getOrders` method for now. Also Cexio offers transactions streaming, but seems this data are too exchange specific, so I think it doesn't worth to move it into new `StreamingPrivateDataService`.

dozd:
Ok then

PR #157, opened 2018-04-04
Add common interface `StreamingPrivateDataService` for private API data

SergeyLukashevich:


xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingExchange.java
```diff
@@ -40,17 +41,17 @@ public StreamingMarketDataService getStreamingMarketDataService() {
...
-        streamingOrderDataService.setApiKey(apiKey);
-        streamingOrderDataService.setApiSecret(apiSecret);
+    @Override
+    public void useCompressedMessages(boolean compressedMessages) {
+        streamingPrivateDataService.useCompressedMessages(compressedMessages);
     }
 
-    public CexioStreamingRawService getStreamingRawService() {
-        return streamingOrderDataService;
+    public void setCredentials(String apiKey, String apiSecret) {
```
declan94: maybe we should just use the apiKey and apiSecret specified in the 'ExchangeSpecification'

PR #172, opened 2018-04-26
Support for trades for the binance exchange

vstavetski:
Implemented trades for the binance exchange.


vstavetski:
Sure, created a pull request #175

PR #182, opened 2018-05-14
add support for okex futures trading

declan94:
Okex futures trading API's endpoint is different from normal trading, which is at wss://real.okex.com:10440/websocket/okexapi.
Here's the API specification: https://github.com/okcoin-okex/API-docs-OKEx.com/blob/master/API-For-Futures-EN/WebSocket%20API%20for%20FUTURES.md

### DIFF
--- a/xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkExStreamingExchange.java
+++ b/xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkExStreamingExchange.java
@@ -10,7 +10,7 @@ public class OkExStreamingExchange extends OkCoinStreamingExchange {
         super(new OkCoinStreamingService(API_URI));
     }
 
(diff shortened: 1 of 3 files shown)


xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkExStreamingExchange.java
```diff
@@ -9,4 +9,8 @@
     public OkExStreamingExchange() {
         super(new OkCoinStreamingService(API_URI));
     }
+
+    public OkExStreamingExchange(String api_url) {
```
dozd: please, use camelCase as it's default for java as well as for this project

declan94: Code has been reformatted according to the standart, thank you for your advice.

PR #192, opened 2018-05-21
fix the bug of multi-contract subscribe

yangguang760:
The key of the orderbooks should contain the information of contract type.
The original code use the currencyPair and different contracts would use the same entry.


Flemingjp:
PR looks fine - @badgerwithagun your call whether it is appropriate to merge this now or later

Issue #189, opened 2018-05-19
Notification of disconnection


declan94:
How to get notified when the wss is disconnectted?

PR #191, opened 2018-05-20
Fix gdax disconnect NPE/Bitmex execution reports/bug-fix rollup

badgerwithagun:
This also rolls in a rebased version of #141 which looked related enough to do at the same time.

Also includes:

* "L1" support for execution reports on BitMex.  This is in-line with the approach for GDAX on #160, just missing the L2 support that #160 adds on top.  I'm happy with this - I am implementing Binance on the same basis (L1 support first, then work out how to create L2-generic support which combines GDAX and BitMex under the same API later).  However, I don't use my bot code on BitMex so I can't easily test that it works.
* Bitfinex proxy support
* Various updated dependencies
* Performance improvements (mainly ensuring we re-use `ObjectMapper` instances and use them more efficiently (use `treeToValue()`)
* Fix for non-3-character tickers on Binance
* API support and support on Bitmex for testnet/sandbox - I'll be adding support for GDAX as soon as this is merged.
* Reconnect/connection success events for some exchanges

PR #193, opened 2018-05-21
Observable for reconnection failure event

declan94:
Add interface to allow user get notified when the WebSocket is disconnected and failed to reconnect.

For issue #189

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -5,6 +5,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -67,6 +68,7 @@ public abstract class NettyStreamingService<T> {
     private final NioEventLoopGroup eventLoopGroup;
     protected Map<String, Subscription> channels = new ConcurrentHashMap<>();
     private boolean compressedMessages = false;
+    private List<ObservableEmitter<Throwable>> reconnFailEmitters = new LinkedList<>();
 
     public NettyStreamingService(String apiUrl) {
         this(apiUrl, 65536);
@@ -225,6 +227,10 @@ public abstract class NettyStreamingService<T> {
         }
     }
 
+    public Observable<Throwable> subscribeReconnectFailure() {
+        return Observable.<Throwable>create(observableEmitter -> reconnFailEmitters.add(observableEmitter));
+    }
+
     public Observable<T> subscribeChannel(String channelName, Object... args) {
         final String channelId = getSubscriptionUniqueId(channelName, args);
         LOG.info("Subscribing to channel {}", channelId);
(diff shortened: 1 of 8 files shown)


service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -225,6 +227,10 @@ public void sendMessage(String message) {
         }
     }
 
+    public Observable<Throwable> subscribeReconnectFailure() {
+        return Observable.<Throwable>create(observableEmitter -> reconnFailEmitters.add(observableEmitter));
```
badgerwithagun: Probably more reactive-standard here to have a single `Flowable` to which we emit the reconnect failures, set up with `.share()`. This method would then just subscribe to that as an `Observable`.

I'm no Rx expert, but am thinking something like this:

```
private final AtomicReference<FlowableEmitter<Throwable>> emitter = new AtomicReference<>();
private final Flowable<Throwable> reconnectFailures = Flowable.create(
    (FlowableEmitter<Throwable> e) -> emitter.set(e.serialize()),
    BackpressureStrategy.MISSING
  ).share();
```

Then:

```
public Observable<Throwable> subscribeReconnectFailure() {
  return reconnectFailures;
}
```

And:

```
 .doOnError(t -> {
  LOG.warn("Problem with reconnect", t);
  emitter.onNext(t);
})
```

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -67,6 +68,7 @@ public Subscription(ObservableEmitter<T> emitter, String channelName, Object[] a
     private final NioEventLoopGroup eventLoopGroup;
     protected Map<String, Subscription> channels = new ConcurrentHashMap<>();
     private boolean compressedMessages = false;
+    private List<ObservableEmitter<Throwable>> reconnFailEmitters = new LinkedList<>();
```
badgerwithagun: My comment below makes this moot, but you should use a concurrency-safe list implementation like `LinkedBlockingQueue`, otherwise you're likely to get `ConcurrentModificationException` if the list is modified while it is being iterated.

xchange-stream-core/src/main/java/info/bitrich/xchangestream/core/StreamingExchange.java
```diff
@@ -26,6 +28,16 @@
...
     boolean isAlive();
 
+    /**
+     * Observable for reconnection failure event.
+     * When this happens, it usually indicates that the server or the network is down.
+     *
+     * @return Observable with the exception during reconnection.
+     */
+    default Observable<Throwable> reconnectFailure() {
+        throw new NotYetImplementedForExchangeException();
```
badgerwithagun: Maybe worth just adding a comment here saying that any `NettyStreamingService` backed exchanges can easily implement this method just by calling `subscribeReconnectFailure()`.

Or implement them all of course ;)

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
declan94: Thank you four your advice, which is great.  I had no experience of Rx at all so the previous implementation is not good. I have modified following your suggestions.

badgerwithagun:
IIRC, there is similar functionality in #191 which is now merged.

badgerwithagun:
@declan94 I think a lot of this may have already got rolled up in other PRs. Would you mind checking how much still needs merging?

badgerwithagun:
Closing this in the absence of progress. Please re-open if necessary.

PR #194, opened 2018-05-24
send ping message to server every 15 sec

declan94:
Hearing from the developer team of Okex, the ws connection would disconnect every 5 minutes if the client doesn't send message {"event": "ping"}.


xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinStreamingService.java
```diff
@@ -4,16 +4,44 @@
...
 
+    @Override
+    public Completable connect() {
+        Completable conn = super.connect();
+        return conn.andThen((CompletableSource)(completable) -> {
+            try {
+                if (pingPongTimer != null) {
+                    pingPongTimer.cancel();
+                }
+                pingPongTimer = new Timer("OkexPingPong", false);
```
dozd: Usage of reactive timer would be nicer here

declan94: Has been changed in the newest commit.

Also, the ping-pong subscription will be disposed now when the ws is disconnected.

PR #195, opened 2018-05-24
handle error of unexpected channel

declan94:
Improve robustness of okcoin streaming service.


service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -295,6 +295,10 @@ protected void handleChannelMessage(String channel, T message) {
     }
 
     protected void handleChannelError(String channel, Throwable t) {
+        if (!channel.contains(channel)) {
```
atomh3art: Shouldn't this be channels.containsKey(channel) ?

badgerwithagun: I think so, yes. Fixed under #191, which is awaiting review.

PR #197, opened 2018-05-28
Initialise binance streaming order book from the fetch api

badgerwithagun:
As detailed at https://github.com/binance-exchange/binance-official-api-docs/blob/master/web-socket-streams.md, you can't start from an empty order book when using the incremental depth API; you need to call the relevant endpoint first. Without this fix, you only get an order book with the most recent changes.

Not 100% convinced about the approach.  This is what I did to resolve an immediate problem but the error handling is suspect in the extreme (I went for the "please don't fall over on my prod box" rather than "fail fast and fail quickly") and the incremental approach used by `OrderBook` isn't quite what Binance recommend (but should probably be OK).

Comments welcome.

### DIFF
--- a/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
+++ b/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
@@ -38,7 +38,7 @@ public class BinanceStreamingExchange extends BinanceExchange implements Streami
 
         ProductSubscription subscriptions = args[0];
         streamingService = createStreamingService(subscriptions);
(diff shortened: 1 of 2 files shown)


jkolobok:
initial approach is correct, but you need to start caching updates and then apply to snapshot only those updates that are after it

jkolobok:
I fixed that properly  in my PR that Bitmex and Binance apis https://github.com/bitrich-info/xchange-stream/pull/185

badgerwithagun:
@jkolobok: testing #185 and it doesn't fix this problem.  Re-opening.

badgerwithagun:
I'm closing this PR as it doesn't solve the problem.

Will try and work on it when I get an opportunity.

PR #202, opened 2018-06-12
prevent concurrent modification exception issue

henry6831:


xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexOrderbook.java
```diff
@@ -27,19 +27,24 @@ public void modify(OrderbookModifiedEvent modifiedEvent) {
         BigDecimal price = modifiedEvent.getPrice();
         BigDecimal volume = modifiedEvent.getVolume();
 
-        side.remove(price);
-        if (volume.compareTo(zero) != 0) {
-            side.put(price, volume);
+        synchronized(side) {
```
badgerwithagun: Have you considered switching `OrderbookModifiedEvent` to use `ConcurrentSkipListMap` rather than `TreeMap`?

badgerwithagun:
@Flemingjp - this is a sensible change, just could make better use of concurrency primitives.  I'd either just tweak it to use `ConcurrentSkipListMap` instead of `TreeMap` and `synchronized` blocks and then merge the result, or just merge it as-is. I'm sure concurrency exceptions are more of an issue than a bit too much lock contention.

PR #209, opened 2018-07-10
[bitfinex] Added some methods of authenticated api

Mikelle:
Added web socket support for getting orders, pre trades and trades, balaces updates


badgerwithagun:
Nice, @Mikelle! 👍 

I'm quite excited by this development; my own application really struggles due to poll limits on exchanges for private authenticated data, so adding streaming support would make a big difference. 

Two questions I think we need to answer:

1. It's a significant decision in principle to extend the scope of xchange-stream to cover authenticated streaming data, so we need @dozd's thoughts on this.
1. Your changes are very BFX focused, rather than introducing a generic API we can extend to cover other exchanges. That's fine, but I think we need to introduce a generic API so that we can introduce other exchanges later. Assuming @dozd is in favour in principle, would you be willing to work together to define a suitable API?

bryantharris:
@badgerwithagun I think the right approach is a two layered approach the same as the REST XChange project that this streaming project is based on.

Basically, layer 0 is exchange specific supporting 1:1 streaming events as to what is supported on the exchange ws API.
Then a layer 1, which would represent generic support, which is implemented using layer 0.  This would be the consistent API to the extent there is commonality shared across all streaming implementations.

You'd typically use the common API but could always drop to the lower level if there were specific event messages you wanted to follow.

With respect to authenticated streaming, I give a resounding thumbs up that it should be supported.  To me the primary advantage of authenticated streaming is real time updates to order status changes without polling.

I get viewing tickers (non authenticated), but the real power seems to be in having the stream be authenticated and receiving private events IMHO.

Mikelle:
@badgerwithagun @bryantharris  So what should I do to make you accept the pull request? Develop new generic API for authenticated streaming?

badgerwithagun:
@Mikelle: thanks for sticking with this.  It would make sense to follow the same approach as @bryantharris has followed on #160 for trades: that is, if we are authenticated, then subscribe to the private trades stream and supply both `Trade`s and `UserTrade`s in the existing trades stream API.  

However, when it comes to orders, we currently lack any generic API, so you will need to define a new one.  I think the `Orderbook` stream API is the closest pattern to follow (maintain a snapshot of the user's order list in memory and keep that updated from the exchange stream).  The level 0 API underlying that would be the actual stream of changes.

I don't have the power to approve change requests though; just helping out.  There are others on here who are closer to the code who would have clearer ideas.

pchertalev:
So we are waiting for approve pool request of #160 for using the same approach for Bitfinex, right?

xchange-bitfinex/src/main/java/info/bitrich/xchangestream/bitfinex/BitfinexStreamingExchange.java
```diff
@@ -57,4 +59,28 @@ public StreamingMarketDataService getStreamingMarketDataService() {
     @Override
     public void useCompressedMessages(boolean compressedMessages) { streamingService.useCompressedMessages(compressedMessages); }
 
+    public Completable connectToAuthenticated() {
```
badgerwithagun: Is it possible to make this connection/authentication transparent, so it occurs automatically if authentication details are available, as we do for GDAX?

xchange-bitfinex/src/main/java/info/bitrich/xchangestream/bitfinex/BitfinexStreamingExchange.java
```diff
@@ -15,15 +15,17 @@
 
     private final BitfinexStreamingService streamingService;
     private BitfinexStreamingMarketDataService streamingMarketDataService;
+    private BitfinexStreamingRawService streamingAuthenticatedDataService;
 
     public BitfinexStreamingExchange() {
         this.streamingService = new BitfinexStreamingService(API_URI);
+        this.streamingAuthenticatedDataService = new BitfinexStreamingRawService(API_URI);
```
badgerwithagun: Do we need to have a second connection to BFX for the authenticated stream, or is it possible to just authenticate the existing connection?

xchange-bitfinex/src/main/java/info/bitrich/xchangestream/bitfinex/BitfinexStreamingRawService.java
```diff
@@ -0,0 +1,303 @@
...
+
+    private void sendMessage(Object message) {
+        try {
+            sendMessage(objectMapper.writeValueAsString(message));
+        } catch (JsonProcessingException e) {
+            LOG.error("Error creating json message: {}", e.getMessage());
+        }
+    }
+
+    public Observable<BitfinexWebSocketAuthOrder> getAuthenticatedOrders() {
```
badgerwithagun: There's an existing pattern for exposing L1 (exchange specific) methods.  Example: `BinanceStreamingMarketDataService.getRawTicker()`. Could we move these to the `StreamingMarketDataService`?  e.g. `getRawAuthenticatedOrders()`?

xchange-bitfinex/src/main/java/info/bitrich/xchangestream/bitfinex/BitfinexStreamingRawService.java
```diff
@@ -0,0 +1,303 @@
...
+        return subjectOrder.share();
+    }
+
+    public Observable<BitfinexWebSocketAuthPreTrade> getAuthenticatedPreTrades() {
+        return subjectPreTrade.share();
+    }
+
+    public Observable<BitfinexWebSocketAuthTrade> getAuthenticatedTrades() {
```
badgerwithagun: I'd like it if we could do as we do for GDAX, and merge the authenticated trades (as `UserTrade`) with the public trades (as `Trade`s) and return them in the `getTrades()` stream.  However, just implementing the L1 API is valuable on its own, so if you don't fancy doing this, that's fine.  We can raise an issue for it and see if someone else picks it up (I might).

xchange-bitfinex/src/main/java/info/bitrich/xchangestream/bitfinex/BitfinexStreamingRawService.java
```diff
@@ -0,0 +1,303 @@
+package info.bitrich.xchangestream.bitfinex;
```
badgerwithagun: I've not reviewed this code - I suspect it might move around quite a lot based on my other comments.

Will review properly once the structure is right.

xchange-bitfinex/src/test/java/info/bitrich/xchangestream/bitfinex/BitfinexManualAuthExample.java
```diff
@@ -0,0 +1,38 @@
...
+public class BitfinexManualAuthExample {
+    private static final Logger LOG = LoggerFactory.getLogger(BitfinexManualAuthExample.class);
+
+    public static void main(String[] args) {
+        BitfinexStreamingExchange exchange = (BitfinexStreamingExchange) StreamingExchangeFactory.INSTANCE.createExchange(
+            BitfinexStreamingExchange.class.getName());
+        exchange.setCredentials("API-KEY", "API-SECRET");
+
+        exchange.connectToAuthenticated().blockingAwait();
+        exchange.getStreamingAuthenticatedDataService().getAuthenticatedTrades().subscribe(
```
badgerwithagun: SO, to clarify my earlier comments, this would be more "standard" like this:

```
ExchangeSpecification specification = StreamingExchangeFactory.INSTANCE.createExchange(BitfinexStreamingExchange.class.getName()).getDefaultExchangeSpecification();
specification.setApiKey("API-KEY");
specification.setSecretKey("API-SECRET");
BitfinexStreamingExchange exchange = (BitfinexStreamingExchange)StreamingExchangeFactory.INSTANCE.createExchange(specification);

exchange.getStreamingMarketDataService().getRawAuthenticatedTrades().subscribe(
 ...
);
```

So, use the existing authentication details provided by the exchange specification, authenticate by default if the authentication details are available, and use L1 "raw" methods on the `StreamingMarketDataService` itself.

davidjirovec:
Any plans on merging this?

Flemingjp:
@davidjirocev If it's tested and reviewed, then most likely.

davidjirovec:
I'm running this branch a few months :)

Flemingjp:
@Mikelle any chance you could work on the comments @badgerwithagun made?

badgerwithagun:
I'm going to fork this PR and bring it in line with the work I'm doing to commonise support for authenticated streams across multiple exchanges (see the thread on #246 for details).

Once #246 is merged I'll be able to resubmit.

badgerwithagun:
@Mikelle, is there a reason why you used a separate `NettyStreamingService` for the authenticated stream, rather than just sending an `AUTH` on the existing stream?

xchange-bitfinex/src/main/java/info/bitrich/xchangestream/bitfinex/BitfinexStreamingExchange.java
badgerwithagun: According to the BFX documentation, it very much looks like this is the case, so I'm going to try it.

xchange-bitfinex/src/main/java/info/bitrich/xchangestream/bitfinex/BitfinexStreamingExchange.java
badgerwithagun: Done on my uplift fork

xchange-bitfinex/src/test/java/info/bitrich/xchangestream/bitfinex/BitfinexManualAuthExample.java
badgerwithagun: Done on my uplift fork

badgerwithagun:
OK, I have created my modified version as #273. This covers all my review issues.

I guess this one can be closed if everyone's happy with the modifications.

badgerwithagun:
Closing this in the absence of any complaints. Can re-open if necessary.

Please head over to #273.

PR #211, opened 2018-07-25
Multiple null pointer fixes and critical Netty fixes.

roeiba:


badgerwithagun:
Possible cross-over with #191 (the graceful pool shutdown).

xchange-gemini/src/main/java/info/bitrich/xchangestream/gemini/GeminiProductStreamingService.java
```diff
@@ -32,7 +32,7 @@ public void messageHandler(String message) {
         try {
             jsonNode = objectMapper.readTree(message);
         } catch (IOException e) {
-            LOG.error("Error parsing incoming message to JSON: {}", message);
+            LOG.error("Error parsing incoming message to JSON.");
```
jkolobok: printing message caused problem is very useful for debugging

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -285,6 +288,11 @@ protected void handleError(T message, Throwable t) {
 
 
     protected void handleChannelMessage(String channel, T message) {
+    	if (channels.get(channel) == null) {
```
jkolobok: this should not be ignored.. if we receive message in channel we are not subscribed to it's a big problem

roeiba:
Thanks for the review. Will fix and create new commit.

badgerwithagun:
Will do.

Bear in mind that my big PR is mostly not my work. I aggregated two previous large PRs which crossed over heavily so I could merge them and test them. Those PRs got closed. I added fixes for the issues I found and posted it all back up as a single PR once I was happy. If I'd had merge permissions that would have been the end of it.

badgerwithagun:
This and #191 both solve the main issue in `NettyStreamingService`, but slightly differently.

The binance change actually looks like it's more likely to introduce an NPE:

Before:
```
    @Override
    public boolean isAlive() {
        return streamingService!= null && streamingService.isSocketOpen();
    }
```
After:
```
    @Override
    public boolean isAlive() {
    		return streamingService.isAlive(); // no null check
    }
```
Unless there's another null check elsewhere already catching this?

The `BitfinexStreamingService` change isn't one I've spotted before. That looks like it should be merged.

xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
```diff
@@ -67,5 +67,7 @@ public ProductSubscription getProductSubscription() {
         return productSubscription;
     }
 
-
+    public boolean isAlive() {
```
badgerwithagun: This doesn't seem necessary. Just call `isSocketOpen()`

badgerwithagun:
I'm guessing that this PR is now too stale to progress in its current form.  There are still issues with `NettyStreamingService`, but a fresh PR is probably the right way to address them.  Closing, unless anyone feels strongly otherwise.

Issue #170, opened 2018-04-24
BitMEX: Error parsing incoming message to JSON


somguyth:
Is anyone else getting a spam of these log messages when trying to receive BitMEX data? I've been getting this since about 30 mins ago, tried reconnecting/restarting multiple times but no luck.

ERROR  [JsonNettyStreamingService] Error parsing incoming message to JSON: B���|��

hlbkin:
Same thing. This is probably because theyve added compression. Need to find out way to add compression somewhere....

somguyth:
@hlbkin  Where does it say they've added compression?

hlbkin:
@somguyth 
I just know it because I watched messages coming in debug mode. It happened with a couple of exchanges in last couple of month and same thing.

Also I've just fixed everything

You need to override method in BitmexStreamingService and create your custom handler for compression

@ChannelHandler.Sharable
public final class BitmexWebsocketHandler extends WebSocketClientExtensionHandler {

    public static final BitmexWebsocketHandler INSTANCE = new BitmexWebsocketHandler();

    private BitmexWebsocketHandler() {
        super(new DeflateFrameClientExtensionHandshaker(false),
                new DeflateFrameClientExtensionHandshaker(true));
    }

}

    @Override
    protected WebSocketClientExtensionHandler getWebSocketClientExtensionHandler() {
        return BitmexWebsocketHandler.INSTANCE;
    }

somguyth:
@hlbkin 

Thanks, but I'm not sure what you mean by "I deleted new PerMessageDeflateClientExtensionHandshaker() from here"?

I did a grep search of the string "new PerMessageDeflateClientExtensionHandshaker()" and could only find such a usage in the gdax library, but no such reference in the bitmex library.

hlbkin:
This is in NettyStreamingService

somguyth:
@hlbkin 

I can confirm this fixes the problem. Thank you very much.

PR #216, opened 2018-08-22
[binance] Fixed json compression error

Hypeon:
See issue https://github.com/bitrich-info/xchange-stream/issues/170

### DIFF
--- a/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
+++ b/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import info.bitrich.xchangestream.core.ProductSubscription;
 import info.bitrich.xchangestream.service.netty.JsonNettyStreamingService;
+import io.netty.handler.codec.http.websocketx.extensions.WebSocketClientExtensionHandler;
 import io.reactivex.Observable;
 import org.knowm.xchange.currency.CurrencyPair;
 import org.knowm.xchange.exceptions.NotYetImplementedForExchangeException;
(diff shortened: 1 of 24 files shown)


Flemingjp:
@Hypeon any chance of updating this PR so that is compiles?

xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
```diff
@@ -67,5 +68,8 @@ public ProductSubscription getProductSubscription() {
         return productSubscription;
     }
 
-
+    @Override
+    protected WebSocketClientExtensionHandler getWebSocketClientExtensionHandler() {
+        return null;
```
badgerwithagun: Should this be `BinanceWebSocketHandler.INSTANCE`?  Just based on https://github.com/sanderdg/xchange-stream/commit/0e2dd0866f09f8a6cecc0531ca84e264523c81b8 which is the Bitmex version of this fix

Flemingjp: This seems to be the case

badgerwithagun: So is this a mistake in the pull request?

Is the pull request actually needed for Binance?

It does sound like it's needed for Bitmex...

Flemingjp:
@badgerwithagun You're referencing https://github.com/sanderdg/xchange-stream/commit/0e2dd0866f09f8a6cecc0531ca84e264523c81b8 right?

badgerwithagun:
Yep

Flemingjp:
Closing as its unclear what this PR is trying to fix. To maintain the prior fix to BitMex, I have opened #233

Hypeon:
@Flemingjp Sorry, that i haven't answered earlier, but github mails were in spam. 

I wasn't clear, but this error occurs for both Bitmex and Binance module. I have just implemented the same thing for Binance.

PR #218, opened 2018-09-11
[BitMex] Refactoring for performance

thermosym:
Issues on previous implemnentation
- use TreeMap as orderbook, the CPU load is heavy when subscribing on multiple CurrencyPairs
- BitmexOrderbook maintain two Map structure, which is redundant
 
Major changes
- Use ConcurrentHashMap to persist orderbook
- In BitmexOrderbook, combine the two maps into a single map, key is 'id', value is the entity
- Code refactoring on orderbook update logic, make it more clear
- Code refactoring on ObjectMapper usage


badgerwithagun:
Significant overlap with #191.

henryxwong:
Anything I can help with this PR? Code review or resolving the conflict?

badgerwithagun:
I'm assuming that this is too stale to be applied in its current form, so closing. @thermosym , if you are still willing and able, it'd be great if you could submit a fresh, up-to-date PR.  I'll be happy to apply it on the assumption that if it works for you, it's good enough.

PR #221, opened 2018-09-23
Fix Binance websockets getting disconnected every few minutes

badgerwithagun:
Binance [requires](https://support.binance.com/hc/en-us/articles/360004492232-API-Frequently-Asked-Questions-FAQ-#%E2%80%9Cwebsocketdrop%E2%80%9D) that pings on websockets are responded with a pong frame.

I've implemented this for all exchanges, since this should be the [standard behaviour](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#Pings_and_Pongs_The_Heartbeat_of_WebSockets).

I have tested with Binance (fixes the connection drops), GDAX and Bitfinex.


badgerwithagun:
@Flemingjp ha! No - it worked for a while stopping the dropped connections and has recently stopped working on my bot. That's probably why.  I'll try and fix that today.

badgerwithagun:
@Flemingjp there we go, done

PR #222, opened 2018-09-24
Default GDAX order book depth to 100

badgerwithagun:
GDAX/Coinbase Pro returns a complete order book by default, and returns them at such an obscene rate that `ObjectMapper` falls behind on most boxes I've tested.  Two things happen: `OutOfMemoryError`s (if you're running with a limited memory footprint) and socket disconnections.

This default of 100 matches the one coded into `BitfinexStreamingMarketDataService`.


badgerwithagun:
Hmm, hadn't noticed that this doesn't even compile.  `develop` is so wildly out of sync with the version I really use.

If anyone at any point actually feels like merging this, poke me and I'll sort it out.  Can't muster the energy to keep a PR green when they're not getting merged.

badgerwithagun:
This is up to date and can be reviewed/merged

Issue #145, opened 2018-03-21
JsonNettyStreamingService does not shutdown all nio threads correctly and application can't be stopped after disposing and disconnecting because stuck nio thread alive


pchertalev:
I tested several exchanges who use JsonNettyStreamingService like BitfinexManualExample, OkCoinManualExample, ...  etc

No one of these tests can stop after disposing and disconnecting.

I have checked thread dump there is alive nio thread after disconneting:

"nioEventLoopGroup-2-1" #14 prio=10 os_prio=2 tid=0x000000002a637800 nid=0x1860 runnable [0x000000002c16e000]
   java.lang.Thread.State: RUNNABLE
	at sun.nio.ch.WindowsSelectorImpl$SubSelector.poll0(Native Method)
	at sun.nio.ch.WindowsSelectorImpl$SubSelector.poll(WindowsSelectorImpl.java:296)
	at sun.nio.ch.WindowsSelectorImpl$SubSelector.access$400(WindowsSelectorImpl.java:278)
	at sun.nio.ch.WindowsSelectorImpl.doSelect(WindowsSelectorImpl.java:159)
	at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:86)
	- locked <0x00000005c11cf628> (a io.netty.channel.nio.SelectedSelectionKeySet)
	- locked <0x00000005c11d0760> (a java.util.Collections$UnmodifiableSet)
	- locked <0x00000005c11d0650> (a sun.nio.ch.WindowsSelectorImpl)
	at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:97)
	at io.netty.channel.nio.SelectedSelectionKeySetSelector.select(SelectedSelectionKeySetSelector.java:62)
	at io.netty.channel.nio.NioEventLoop.select(NioEventLoop.java:753)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:409)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:886)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)

badgerwithagun:
This may be solved by #191

pchertalev:
solved by #191

Issue #35, opened 2017-11-18
Exchange doesn't disconnect gracefully


jpink:
When I repeat following steps as in example:
1. Connect to GDAX exchange.
2. Subscribe to trades.
3. Check that there really come trades.
4. Dispose all my subscriptions.
5. Disconnect from exchange.
6. Program main loop goes to its end.

But program doesn't stop! There is still threads running somewhere.

I'm using Spring Boot web app and I'm stopping web server via POST /shutdown request. All other services are then shutdown.

This is not big deal. Currently I call System.exit() 15 s after that request. But this is annoying and smells that there is thread leak somewhere.

dozd:
@jpink Yeah, I can reproduce that and it's bug of exchanges based on `NettyStreamingService`.

dcarr45:
Did anyone ever do anything about this? The EventLoopGroup thread is still running I think, but I don't know enough about these things

bryantharris:
This one is pretty annoying, sure I can add a System.exit() call so that I'm not chasing down PIDs but it breaks WAR hot deploy as I need to stop tomcat and restart it to pick up code changes, I should be able to just trigger the deploy from rebuilding.

bryantharris:
This issue also seems to effect Gemini.

badgerwithagun:
Possibly related - I get an NPE whenever I disconnect from GDAX due to a background socket thread:

```
java.lang.NullPointerException: null
	at info.bitrich.xchangestream.service.netty.NettyStreamingService.handleChannelMessage(NettyStreamingService.java:288)
	at info.bitrich.xchangestream.service.netty.NettyStreamingService.handleMessage(NettyStreamingService.java:278)
	at info.bitrich.xchangestream.gdax.GDAXStreamingService.handleMessage(GDAXStreamingService.java:84)
	at info.bitrich.xchangestream.gdax.GDAXStreamingService.handleMessage(GDAXStreamingService.java:21)
	at info.bitrich.xchangestream.service.netty.JsonNettyStreamingService.messageHandler(JsonNettyStreamingService.java:42)
	at info.bitrich.xchangestream.service.netty.WebSocketClientHandler.channelRead0(WebSocketClientHandler.java:79)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
[...]

Issue #224, opened 2018-10-02
Project Continuation


fynnfluegge:
Hello,

I want to ask what is the roadmap for continuation of the project.
I saw there is a maintainer needed. As I currently need as many websocket implemantations for various exchanges as possible, I am open for pushing the project forward.
Currently 12 exchange streams are implemented, but I guess meanwhile there may be some more exchanges providing a websocket API. Correct me if I'm wrong...

davidjirovec:
It is sad to see so many pull requests, which are not merged. Maybe people would contribute even more, if they would know their pull requests will get merged. I understand @dozd might not have time for this project, but this project should go on. What is the best solution? Is there anybody to maintain fork of this project?

Flemingjp:
I don't have the time to maintain a whole fork, but would invest some time in contribution. It's difficult to invest any time when I know this repository is becoming stagnant.

oreonengine:
I may be willing to maintain a fork, but I don't think that I if can do this alone since not that much time...

Flemingjp:
I also think it will help this repository by reviewing the current PRs. @dozd [explained ](https://github.com/bitrich-info/xchange-stream/issues/181#issuecomment-395054903)that they would be happy to merge from then on.

oreonengine:
Yeah the primary goal should review all PRs and merge as many as possible, I think it would help if @dozd would give some chosen people write access and finally synch the develop branch with all forks.

Another point, I am considering to start another project that provides Java Websocket streams for cryptocompare API, anyone interested in this? I think it's good to have various streaming APIs

Flemingjp:
If we can pull through as many PR as possible, and then look towards releasing a new version of XChange-Streams to be inline with the, quite recent, version 4.3.11 of XChange, I think that'd be a good start.

badgerwithagun:
I already maintain a separate development branch for my own purposes, which is `develop` + all my PRs + some other selected PRs that I needed: https://github.com/badgerwithagun/xchange-stream/tree/latestandgreatest

The divergence from `develop` is rapidly getting impossible to manage though. I have no wish to fork someone's project on a permanent basis. I'd much rather see the main project succeed.

For example, #185, #191 and now #225 are all dealing with the same set of problems.  I've managed to combine  #185 and #191 in a way that works, but all it will take is for someone to merge one and not the other, or combine them in a different way, and I will be in a world of pain.

[...]

Issue #190, opened 2018-05-20
Bitfinex tickers do not support multiple subscribers


badgerwithagun:
With multiple subscribers to the same ticker stream, only one receives:

```
import static org.knowm.xchange.currency.CurrencyPair.BTC_USD;
import org.junit.Test;
import info.bitrich.xchangestream.bitfinex.BitfinexStreamingExchange;
import info.bitrich.xchangestream.core.ProductSubscription;
import info.bitrich.xchangestream.core.StreamingExchange;
import info.bitrich.xchangestream.core.StreamingExchangeFactory;
import info.bitrich.xchangestream.core.StreamingMarketDataService;
import io.reactivex.disposables.Disposable;

public class TestReplicateBitfinexIssue {

  @Test
  public void test() throws InterruptedException {
    StreamingExchange exchange = StreamingExchangeFactory.INSTANCE.createExchange(BitfinexStreamingExchange.class.getName());
    exchange.connect(ProductSubscription.create().addTicker(BTC_USD).build()).blockingAwait();
    try {
      StreamingMarketDataService streamingMarketDataService = exchange.getStreamingMarketDataService();

      Disposable subscribe1 = streamingMarketDataService.getTicker(BTC_USD)
        .subscribe(
          t -> System.out.println("Subscriber 1: " + t),
          e -> System.out.println("Error subscriber 1: " + e)
        );
      Disposable subscribe2 = streamingMarketDataService.getTicker(BTC_USD)
        .subscribe(
          t -> System.out.println("Subscriber 2: " + t),
          e -> System.out.println("Error subscriber 1: " + e)
        );
      Thread.sleep(60000);
      subscribe1.dispose();
      subscribe2.dispose();
    } finally {
      exchange.disconnect();
    }
  }
}
```

Output example:
```
Subscriber 1: Ticker [currencyPair=BTC/USD, open=null, last=8344.2, bid=8344.2, ask=8344.3, high=8468, low=8171.5,avg=null, volume=13768.14389861, quoteVolume=null, timestamp=1526810221000, bidSize=null, askSize=null]
Subscriber 1: Ticker [currencyPair=BTC/USD, open=null, last=8344.3, bid=8344.2, ask=8344.3, high=8468, low=8171.5,avg=null, volume=13768.2543546, quoteVolume=null, timestamp=1526810225000, bidSize=null, askSize=null]
```

With others (GDAX and Binance tested) both subscribers get the tickers.

[...]

PR #225, opened 2018-10-02
Defect: NettyStreamingService.disconnect() method does not shutdown a…

pchertalev:
…ll threads - FIXED

+ FIX NullPointerException on first "partial"

Looks like it is fix of bugs: #145 #35

### DIFF
--- a/xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinStreamingService.java
+++ b/xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinStreamingService.java
@@ -4,14 +4,44 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import info.bitrich.xchangestream.okcoin.dto.WebSocketMessage;
 import info.bitrich.xchangestream.service.netty.JsonNettyStreamingService;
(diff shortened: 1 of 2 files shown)


Flemingjp:
Recommdation would be to close this PR?

badgerwithagun:
I'm going to try merging this locally to see the differences in place.   WIll report back

badgerwithagun:
OK, so it's a mixed bag.

 - The changes to the *Example.java files should be merged.
 - The change to `BitmexStreamingMarketDataService` is already implemented in #191.  The version in #191 is slightly safer (immutable) but they do the same thing.
 - The changes to `NettyStreamingService` are sensible too, and also again implemented in #191.  However, I'm unsure about the differences.  #191 switches to a 2-thread event loop group and I don't know why (I didn't write it).  There's usually a reason for this stuff.  There's also extra logging in #191.

This is a good PR.  If it's merged before #191 it would be beneficial on its own.  However, it would be a pain to fix up #191 again so it merges, and that would screw up the cleanup proposal I put in #224. 

My preference would be to get  #191 merged first, then follow up with this when we can make a choice about the event group thread count question.

Flemingjp:
Ok, this should be put on hold until we can merge #191

pchertalev:
No problemo. I will resolve conflicts after merge #190

PR #227, opened 2018-10-08
[Binance] Orderbook initial snapshot

### LINKED ISSUE #86: Binance Orderbook Streaming implementation doesn't take the initial snapshot
Binance API documentation gives instructions how to manage local orderbook correctly:
https://github.com/binance-exchange/binance-official-api-docs/blob/master/web-socket-streams.md
## How to manage a local order book correctly
1. Open a stream to **wss://stream.binance.com:9443/ws/bnbbtc@depth**
2. Buffer the events you receive from the stream
3. Get a depth snapshot from **https://www.binance.com/api/v1/depth?symbol=BNBBTC&limit=1000"
4. Drop any event where `u` is <= `lastUpdateId` in the snapshot
5. The first processed should have `U` <= `lastUpdateId`+1 **AND** `u` >= `lastUpdateId`+1
6. While listening to the stream, each new event's `U` should be equal to the previous event's `u`+1
7. The data in each event is the **absolute** quantity for a price level
8. If the quantity is 0, **remove** the price level
9. Receiving an event that removes a price level that is not in your local order book can happen and is normal.

As of xchange-stream 4.3.1 it doesn't follow the recommendation by Binance API Docs. As you can see the recommendation is to get the snapshot of full orderbook on 3rd step, and then apply updates on it. The current implementation doesn't take the snapshot. As a result the local orderbook doesn't reflect the reality of Binance orderbook on serverside (though it slowly gets closer over time).

### LINKED ISSUE #203: [Binance] Order book gets out of sync after a while
I've noticed that an order book snapshot gets out of sync over time (even including #197).  We're definitely not using _quite_ the same approach described in Binance's docs for accurately syncing via the streaming API.  The solution is probably to follow that more slavishly and/or to do periodic refreshes from the pull API.

I'll be looking into this properly when I get a chance. Just recording it here for reference.

badgerwithagun:
Resolves #86.  Explanation given on there.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
     </repositories>
 
     <properties>
-        <xchange.version>4.3.5</xchange.version>
+        <xchange.version>4.3.9</xchange.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <maven.compiler.source>1.8</maven.compiler.source>
--- a/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
+++ b/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
@@ -5,6 +5,7 @@ import info.bitrich.xchangestream.core.StreamingExchange;
 import info.bitrich.xchangestream.core.StreamingMarketDataService;
 import io.reactivex.Completable;
 import org.knowm.xchange.binance.BinanceExchange;
+import org.knowm.xchange.binance.service.BinanceMarketDataService;
 import org.knowm.xchange.currency.CurrencyPair;
 
 import java.util.List;
@@ -38,7 +39,7 @@ public class BinanceStreamingExchange extends BinanceExchange implements Streami
 
         ProductSubscription subscriptions = args[0];
         streamingService = createStreamingService(subscriptions);
-        streamingMarketDataService = new BinanceStreamingMarketDataService(streamingService);
+        streamingMarketDataService = new BinanceStreamingMarketDataService(streamingService, (BinanceMarketDataService) marketDataService);
         return streamingService.connect()
                 .doOnComplete(() -> streamingMarketDataService.openSubscriptions(subscriptions));
     }
--- a/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
+++ b/xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
@@ -13,8 +13,6 @@ import info.bitrich.xchangestream.core.ProductSubscription;
 import info.bitrich.xchangestream.core.StreamingMarketDataService;
 import io.reactivex.Observable;
 import io.reactivex.functions.Consumer;
-import io.reactivex.observables.ConnectableObservable;
-
 import org.knowm.xchange.binance.BinanceAdapters;
 import org.knowm.xchange.binance.dto.marketdata.BinanceOrderbook;
 import org.knowm.xchange.binance.dto.marketdata.BinanceTicker24h;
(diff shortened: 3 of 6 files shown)


davidjirovec:
I am afraid there is some memory leak. When I modify manual example like following
```
public static void main(String[] args) throws InterruptedException {
    final ExchangeSpecification exchangeSpecification = new ExchangeSpecification(BinanceStreamingExchange.class);
    exchangeSpecification.setShouldLoadRemoteMetaData(true);
    StreamingExchange exchange = StreamingExchangeFactory.INSTANCE.createExchange(exchangeSpecification);
    ProductSubscription subscription = exchange.getExchangeSymbols().stream().limit(50).reduce(ProductSubscription.create(), ProductSubscription.ProductSubscriptionBuilder::addOrderbook, (productSubscriptionBuilder, productSubscriptionBuilder2) -> {
        throw new UnsupportedOperationException();
    }).build();
    exchange.connect(subscription).blockingAwait();
    Thread.sleep(MAX_VALUE);
}
```
After 45 minutes memory usage look like this, pattern keeps growing like this even when "Perform GC" is clicked and eventually it would run out of memory.
heap

badgerwithagun:
@davidjirovec : drat. Thanks for that report. Really useful.  Did you confirm that this wasn't occurring beforehand?

I'll look into this in the next couple of days.

badgerwithagun:
Just checked my production app and it has been quietly cycling every few hours due to OOM errors, but only slightly more often than it always has.  There's definitely already a memory leak somewhere.  If this has made it worse, that provides some clues perhaps...

davidjirovec:
I didn't use xchange-stream binance until this PR. You used some previous version even when it gave you OOM? :)

badgerwithagun:
Background thread issuing a System.exit() every few hours and the container restarts itself.

Filth I know, but I have had more important things to worry about. Memory leaks are somewhere below Urgent.

I will look at this properly now tho.

badgerwithagun:
Hmmm.  I'm not getting the same thing.

image

badgerwithagun:
Confirmed.  This memory leak goes away when #191 is merged.

@davidjirovec , if you'd like to confirm yourself, compare the results of running your test against this branch (`binance-orderbook-initial-snapshot`) and `develop-fork`, both on my fork. The only real material differences between the branches are that the latter has some slightly newer dependencies (but I've tested and they aren't having any effect) and that the latter contains #191, which cleans up connection/reconnection issues.

`binance-orderbook-initial-snapshot`:

image

`develop-fork`:

image

In short, I think this PR is OK.

badgerwithagun:
> Thanks, I will test it. `develop-fork` should be handling orderbooks correctly? Because it to me it seems the code is missing. But I've found your branch `develop-fork-develop` (wow :) ) and to me it seems it contains both fixes for connection issues and orderbook handling? But maybe I am not right, it is complicated :)

Ignore `develop-fork-develop` - that's my own personal workspace :).

 - `develop-fork` contains this fix *plus* #191 (plus a few other things, but I've already tested with and without those changes, and they don't affect this result).
 - `binance-orderbook-initial-snapshot` contains only this fix.

You should find that `develop-fork` produces a flat memory profile, but `binance-orderbook-initial-snapshot` shows the leak.  This shows (to me at least) that the leak has nothing to do with this PR but is fixed by #191.

badgerwithagun:
Sorry @davidjirovec, you were quite right, `develop-fork` *was* missing this PR.  So many forks.  I've pushed the update now, but don't run your tests yet.  I'm just double-checking.  I'm certain I tested this on `develop-fork-develop` *too*, but now I'm doubting myself.

badgerwithagun:
I was indeed confusing myself. I've now replicated this on `develop-fork` too.

OK, I will go back to trying to work out what the cause is.

badgerwithagun:
@davidjirovec - found it, fixed

davidjirovec:
Tested myself, looks good!

badgerwithagun:
@davidjirovec, are you able to post an accepting review (see the **Review Required** section below)?  Then we can get it merged.

davidjirovec:
Says "At least 1 approving review is required by reviewers with write access." :-/

Flemingjp:
@davidjirovec It just needs someone with write access to also accept - I can sort this out for you

Issue #205, opened 2018-06-21
[GDAX] Change endpoint to ws-feed.pro.coinbase.com


malkusch:
GDAX users just received an email that the old GDAX endpoint became deprecated and users should migrate to the new Coinbase Pro endpoint before the end of the year:

> GDAX APIs will continue to function until December 31, 2018. We encourage Coinbase Pro users to begin migrating their API usage away from api.gdax.com to api.pro.coinbase.com.

The [new documentation](https://docs.pro.coinbase.com/#websocket-feed) points to ws-feed.pro.coinbase.com.

It seems that they want to rebrand GDAX into Coinbase Pro. Maybe it's also appropriate to rename the complete package as well, as GDAX won't have any meaning in the future.

See also https://github.com/knowm/XChange/issues/2668.

badgerwithagun:
Fixing this, along with the fact that we're currently hard-coded to use the production wss endpoint, which doesn't work if the exchange specification is set to the sandbox endpoint.

Flemingjp:
@badgerwithagun Is there a PR (either open or closed) relating to this issue?

badgerwithagun:
No, was waiting for #160 to get cleared. I can probably raise this now. Will do.

badgerwithagun:
RIght, @Flemingjp, the change I have on my fork does technically resolve this, but unfortunately I don't think it's the right solution.

XChange has implemented Coinbase Pro as a new exchange and deprecated the GDAX implementation.  The _right_ thing to do is to copy the GDAX code, rename it, deprecate the old code and then add the sandbox support there (using the sandbox API added on #191).

I don't realistically think we can do that until we've got the big boys merged and moved up to 4.3.11.

Flemingjp:
It maybe a good idea to deprecate the GDAX Exchange and Service classes before the release of 4.3.11 as a heads up to those using XChange-Streams in production. Just so they're aware of an upcoming change.

badgerwithagun:
Agreed

badgerwithagun:
Well, we never got around to any of the above, 4.3.12 of XChange came and went, it's now XChange 4.3.13 and the GDAX chasses in XChange have been completely removed.

I've done the same in this library and bumped up to XChange 4.3.13 in #261.

badgerwithagun:
This was done a while ago. Closing.

PR #237, opened 2018-10-11
[GDAX]  Add dynamic URL support

theo1985:
Added support to change URL to support Coinbase Pro/Prime


Flemingjp:
This is a good temporary fix, but ultimately #205 will be the way to go with CoinBasePro/Prime. Unless this is a desperately needed patch, I'm tempted to hold off on this PR.

Flemingjp:
Will reopen if this becomes absolutely necessary

PR #245, opened 2018-10-25
Update Jackson and Gson dependencies

badgerwithagun:
I've been running with these for some time.

### DIFF
--- a/service-netty/pom.xml
+++ b/service-netty/pom.xml
@@ -26,7 +26,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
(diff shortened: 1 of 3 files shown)


Flemingjp:
Looking at the change logs there seems to be no glaring issues with updating these. Just a few notes for future reference.

> **Jackson 2.9 Change Log**
As per [databind#219](https://github.com/FasterXML/jackson-databind/issues/219) `java.sql.Date` will finally use same "timestamp-or-String" determination as `java.util.Date` and `java.util.Calendar`. This means that with vanilla, unchanged settings, values will be serialized as numeric timestamps. Note that the default String serialization will still default to `java.sql.Date.String()`, and not to default formatting `java.util.Date` uses.

Issue #230, opened 2018-10-09
[GDAX] UserTrades and Trades duplicate each other when using authenticated stream


badgerwithagun:
As discussed in review of #160, if a user trade executes and the stream is authenticated, we return a `UserTrade` for that trade (from the authenticated `user` channel), but we _also_ return the public `Trade` (from the public `fills` channel).

*Do we consider this a bug*?  The only way I can think of to de-duplicate the two, giving preference to the `UserTrade`, is to delay the `fills` stream by _just enough_ that we can assume that the `UserTrade` will arrive first, then skip any IDs we've seen recently.

Example implementation of `GDAXStreamingMarketDataService.getTrades()`:

```
// To avoid duplication of user and public trades, we delay the public trades by one second
// and skip any trade ids which have recently been streamed as user trades. This isn't
// perfect, of course...
Cache<String, ? super Trade> recentTrades = CacheBuilder.newBuilder().maximumSize(50).build();
return publicTrades
  .concatMap(t -> Observable.just(t).delay(1, TimeUnit.SECONDS))
  .skipWhile(t -> recentTrades.asMap().containsKey(t.getId()))
  .mergeWith(
    getUserTrades()
      .filter(t -> t.getCurrencyPair().equals(currencyPair))
      .doOnNext(t -> recentTrades.put(t.getId(), t))
  );
```

Flemingjp:
It is odd behaviour. My gut reaction would be to avoid any delays and return the current duplicated stream - this is something the user can resolve but in reality its a problem that *every* user will have to solve, and will cause much confusion.

Is there any way to check if authenticated and then just use one stream of trades or the other - avoid merging them at all?

badgerwithagun:
It certainly wouldnt work for me - I require both public and private trades. Some exchanges really don't like you opening multiple sockets; their rate limiting can be quite aggressive in that regard.

I think it'd be stretching the contract of the current API too.

One argument is that maybe backing away from mixing authenticated and unauthenticated is the right solution.

bryantharris:
Is this the result of the implementation or a result of the way gdax sends events based on the various types you've subscribed to?

If gdax sends the trade out twice for two different subscription types then I'm more inclined to say the bot/client should expect this behavior.

badgerwithagun:
This may be true, but if it is, I think it would make more sense to split the streams and have a new `getUserTrades()` method in the API.

fynnfluegge:
[...]

PR #244, opened 2018-10-25
[BitMex] Restore the Bitmex Execution reports code which was cut out of #191.

badgerwithagun:
#191 originally contained Bitmex execution report support.  It had heavy overlap with some other urgent PRs, so maintaining a "roll up" of these PRs (which I did in #191) was the only practical way of managing the conflicts during the months when PRs weren't getting applied.

Now, however, things are moving and we are slowly clearing the backlog of urgent PRs.

This Bitmex functionality was mixed in with various other useful performance tweaks and bug fixes which we wanted to get merged urgently.  However, we have been short of Bitmex testers/reviewers so this functionality was extracted from that PR with the intention of reapplying it later:

I hope I have correctly credited the original author in this squashed commit.

Note that this doesn't compile.  If someone who uses Bitmex could pick this up and get it compiling, that'd be great.

PR #246, opened 2018-10-28
[Binance] Support for execution reports / UserTrades / fills

badgerwithagun:
This follows the same pattern as #160 did for GDAX.  Execution reports are now fetched from Binance if you connect with an API key.  That makes the following available:

# L2 (Generic) API

`StreamingMarketDataService.getTrades(...)`: now returns both `UserTrade` and `Trade` instances.  As with the existing solution for GDAX, these are duplicated (see #230 

# L1 API

The following are available if you cast your `StreamingMarketDataService` to `BinanceStreamingMarketDataService`:

* `BinanceStreamingMarketDataService.getRawExecutionReports()`: full raw details of all trades on all currencies, order creations and order fills.  This is the raw Binance API (and is thus subject to change if the Binance API itself changes).
* `BinanceStreamingMarketDataService.getUserTrades()`: Execution reports filtered down to just trades and converted to `UserTrade`.
* `BinanceStreamingMarketDataService.getUserTrades(CurrencyPair)`: As above, but filtered to a specific currency.

I think there's an argument for supporting the latter two as part of the generic API.  Thoughts?


badgerwithagun:
I've been running with this for some time, with no problems.

Anyone got a review in them?

badgerwithagun:
Still no reviews for this?

badgerwithagun:
Anyone?

badgerwithagun:
@Flemingjp , I can't wait any longer and am now turning this into generic support across Binance, Bitfinex and Coinbase Pro, with more to come.   That code is now ready to push but I don't want to complicate this PR it I don't have to; I'd rather submit the next bit as a separate PR.

If this isn't merged soon it's going to become another monster PR with useful features that never gets merged because it's too big for anyone to review, and we're going to be frozen regarding Binance in the same way as we are with Bitmex.

I am probably going to start looking at Bitmex next (#267, #244 and ensuring consistency with this), but I can't do anything about that either until this is merged.

I suggest either you and I make a decision on this now, or we call upon a couple of other regular contributors to agree on it, because if not, I can't afford to keep spending the time balancing two divergent code sets.

badgerwithagun:
See #274 for details of the generic API I'm working on which builds on this.

Flemingjp:
@badgerwithagun  Apologies for the delay - are you still wanting this merged to help support your L2 api additions?

badgerwithagun:
Yes please! Once this and the Bitfinex one are in, I'll be able to publicise my dev branch so I can get feedback on progress.

badgerwithagun:
Thanks @Flemingjp .

PR #250, opened 2018-10-29
Fix of the netty streaming socket leak problem + some small fixes

pchertalev:
Fix of the socket leak problem for all implemtations which use netty streaming + some small fixes

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -179,7 +179,7 @@ public abstract class NettyStreamingService<T> {
                                 }
 
                                 handlers.add(handler);
-                                p.addLast(handlers.toArray(new ChannelHandler[handlers.size()]));
+                                p.addLast(handlers.toArray(new ChannelHandler[0]));
                             }
                         });
 
@@ -210,11 +210,7 @@ public abstract class NettyStreamingService<T> {
             LOG.warn("Resubscribing channels");
             resubscribeChannels();
 
-            connectionSuccessEmitters.forEach(emitter -> {
-                    emitter.onNext(new Object());
-                }
-            );
-
+            connectionSuccessEmitters.forEach(emitter -> emitter.onNext(new Object()));
         });
     }
 
@@ -226,6 +222,8 @@ public abstract class NettyStreamingService<T> {
                 isManualDisconnect = false;
             }
         });
+        // shutdown sockets after disconnect for avoiding sockets leak
+        eventLoopGroup.shutdownGracefully(2, 30, TimeUnit.SECONDS);
         completable.onError(t);
     }
 
@@ -290,11 +288,11 @@ public abstract class NettyStreamingService<T> {
     }
 
     public Observable<Throwable> subscribeReconnectFailure() {
-        return Observable.<Throwable>create(observableEmitter -> reconnFailEmitters.add(observableEmitter));
+        return Observable.create(reconnFailEmitters::add);
     }
 
     public Observable<Object> subscribeConnectionSuccess() {
-        return Observable.<Object>create(e -> connectionSuccessEmitters.add(e));
+        return Observable.create(connectionSuccessEmitters::add);
     }
 
     public Observable<T> subscribeChannel(String channelName, Object... args) {
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -221,9 +221,9 @@ public abstract class NettyStreamingService<T> {
             if(f.isSuccess()) {
                 isManualDisconnect = false;
             }
+            // shutdown sockets after disconnect for avoiding sockets leak
+            eventLoopGroup.shutdownGracefully(2, 30, TimeUnit.SECONDS);
         });
-        // shutdown sockets after disconnect for avoiding sockets leak
-        eventLoopGroup.shutdownGracefully(2, 30, TimeUnit.SECONDS);
         completable.onError(t);
     }
 
--- a/xchange-bitfinex/src/test/java/info/bitrich/xchangestream/bitfinex/BitfinexManualExample.java
+++ b/xchange-bitfinex/src/test/java/info/bitrich/xchangestream/bitfinex/BitfinexManualExample.java
@@ -60,14 +60,13 @@ public class BitfinexManualExample {
                     LOG.error("ERROR in getting trade: ", throwable);
                 });
 
+        Thread.sleep(10000);
+
         tickerSubscriber.dispose();
         tradesSubscriber.dispose();
         orderBookSubscriber.dispose();
 
         LOG.info("disconnecting...");
         exchange.disconnect().subscribe(() -> LOG.info("disconnected"));
-
-        Thread.sleep(10000);
-
     }
 }
--- a/xchange-bitfinex/src/test/java/info/bitrich/xchangestream/bitfinex/BitfinexManualExample.java
+++ b/xchange-bitfinex/src/test/java/info/bitrich/xchangestream/bitfinex/BitfinexManualExample.java
@@ -2,8 +2,11 @@ package info.bitrich.xchangestream.bitfinex;
 
 import info.bitrich.xchangestream.core.StreamingExchange;
 import info.bitrich.xchangestream.core.StreamingExchangeFactory;
+import io.reactivex.Observable;
+import io.reactivex.disposables.Disposable;
 import org.knowm.xchange.ExchangeSpecification;
 import org.knowm.xchange.currency.CurrencyPair;
+import org.knowm.xchange.dto.marketdata.OrderBook;
 import org.knowm.xchange.utils.CertHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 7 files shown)


pchertalev:
Guys, please review my pool request

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -226,6 +222,8 @@ protected void handleError(CompletableEmitter completable, Throwable t) {
                 isManualDisconnect = false;
             }
         });
+        // shutdown sockets after disconnect for avoiding sockets leak
+        eventLoopGroup.shutdownGracefully(2, 30, TimeUnit.SECONDS);
```
badgerwithagun: I wonder if there might be any potential race conditions from the fact that this is an asynchronous call. `completable.onError(t)` will fire immediately and the shutdown will occur in the background, possibly while a retry reconnect starts.

I am not sure.  Anyone else?

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -179,7 +179,7 @@ protected void initChannel(SocketChannel ch) {
                                 }
 
                                 handlers.add(handler);
-                                p.addLast(handlers.toArray(new ChannelHandler[handlers.size()]));
+                                p.addLast(handlers.toArray(new ChannelHandler[0]));
```
badgerwithagun: Why this change?  The existing code was fractionally more efficient than this (good discussion on it here: https://stackoverflow.com/questions/3293946/the-easiest-way-to-transform-collection-to-array)

Flemingjp:
@pchertalev someone will, but you need patience, there a very few active developers at the moment

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
pchertalev: JetBrains Idea said that better to use empty array in ArrayTo function. :)

pchertalev: reference
intellij
Call to 'toArray()' with pre-sized array argument 'new PathMatcher[collection.size()]' less... (Ctrl+F1)
There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow.
However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version.
Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array,
if the collection was concurrently shrunk during the operation.
This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

pchertalev: https://medium.com/@hanru.yeh/collection-toarray-new-t-0-will-be-faster-than-collection-toarray-new-t-size-5b129fabfa88

badgerwithagun: I am both convinced and educated :)

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
badgerwithagun: I agree it fixes the leak, for sure.

I think I'm convinced on the async too. Good! Thanks for the explanations.

badgerwithagun:
I'm going to merge this into my code tonight and test it against my app for a day or two, which connects to a lot of exchanges for a long time.

pchertalev:
> I'm going to merge this into my code tonight and test it against my app for a day or two, which connects to a lot of exchanges for a long time.

Thanks

badgerwithagun:
It's been in since yesterday and I've spotted a couple of connectivity issues. I really don't think it's related to this, but I want to be sure so am running it a bit longer and if they occur again, I'll trying rolling out some other recent changes to see if they are the cause.

badgerwithagun:
Still testing.  @pchertalev : if you use jitpack.io you can resolve your own code from Maven, which might unblock you in the meantime.

pchertalev:
I have found problem related with my fix.

webSocketChannel.disconnect();  // async call
eventLoopGroup.shutdownGracefully(2, 30, TimeUnit.SECONDS); 

Method webSocketChannel.disconnect() is async as well and eventLoopGroup.shutdownGracefully() can be completed before webSocketChannel.disconnect() - in this case I got infinit loop with exception:

`
2018-11-01 12:33:40.170 ERROR [RxComputationThreadPool-10] DefaultPromise.rejectedExecution - Failed to submit a listener notification task. Event loop shut down?
java.util.concurrent.RejectedExecutionException: event executor terminated
	at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:844)
	at io.netty.util.concurrent.SingleThreadEventExecutor.offerTask(SingleThreadEventExecutor.java:328)
	at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:321)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:769)
	at io.netty.util.concurrent.DefaultPromise.safeExecute(DefaultPromise.java:765)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:433)
	at io.netty.util.concurrent.DefaultPromise.setFailure(DefaultPromise.java:113)
	at io.netty.channel.DefaultChannelPromise.setFailure(DefaultChannelPromise.java:89)
	at io.netty.channel.AbstractChannelHandlerContext.safeExecute(AbstractChannelHandlerContext.java:1010)
	at io.netty.channel.AbstractChannelHandlerContext.disconnect(AbstractChannelHandlerContext.java:572)
	at io.netty.channel.AbstractChannelHandlerContext.disconnect(AbstractChannelHandlerContext.java:460)
	at io.netty.channel.DefaultChannelPipeline.disconnect(DefaultChannelPipeline.java:978)
	at io.netty.channel.AbstractChannel.disconnect(AbstractChannel.java:233)
	at info.bitrich.xchangestream.service.netty.NettyStreamingService.handleError(NettyStreamingService.java:225)
	at info.bitrich.xchangestream.service.netty.NettyStreamingService.lambda$connect$2(NettyStreamingService.java:204)
	at io.reactivex.internal.operators.completable.CompletableCreate.subscribeActual(CompletableCreate.java:39)
	at io.reactivex.Completable.subscribe(Completable.java:1635)
	at io.reactivex.internal.operators.completable.CompletablePeek.subscribeActual(CompletablePeek.java:51)
	at io.reactivex.Completable.subscribe(Completable.java:1635)
	at io.reactivex.internal.operators.completable.CompletableToFlowable.subscribeActual(CompletableToFlowable.java:32)
	at io.reactivex.Flowable.subscribe(Flowable.java:12978)
	at io.reactivex.Flowable.subscribe(Flowable.java:12924)
	at io.reactivex.internal.operators.flowable.FlowableRepeatWhen$WhenReceiver.onNext(FlowableRepeatWhen.java:101)
	at io.reactivex.internal.operators.flowable.FlowableFlatMap$MergeSubscriber.tryEmit(FlowableFlatMap.java:282)
	at io.reactivex.internal.operators.flowable.FlowableFlatMap$InnerSubscriber.onNext(FlowableFlatMap.java:657)
	at io.reactivex.internal.operators.flowable.FlowableTimer$TimerSubscriber.run(FlowableTimer.java:76)
	at io.reactivex.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:38)
	at io.reactivex.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:26)
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266)
	at java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
`

badgerwithagun:
That could well be what was causing my application to fall over.  This morning I reverted your fix out of my application to see if it was more stable, and it appears to have been stable since then.  I was going to properly look at the logs later.

badgerwithagun:
So something like this?

```
disconnect.addListener(f -> {
  if(f.isSuccess()) {	
    isManualDisconnect = false;
  }
  eventLoopGroup.shutdownGracefully(2, 30, TimeUnit.SECONDS);
});
```

pchertalev:
> So something like this?
> 
> ```
> disconnect.addListener(f -> {
>   if(f.isSuccess()) {	
>     isManualDisconnect = false;
>   }
>   eventLoopGroup.shutdownGracefully(2, 30, TimeUnit.SECONDS);
> });
> ```
I think somethink like this
`        
        ChannelFuture disconnect = webSocketChannel.disconnect();
        disconnect.addListener(f -> {
            if(f.isSuccess()) {
                isManualDisconnect = false;
            }
        });
        disconnect.await(retryDuration.toMillis());   /// wait here until disconnect complete
        eventLoopGroup.shutdownGracefully(2, 30, TimeUnit.SECONDS);
`

pchertalev:
I found that problem after infinite looping on such exception:

2018-11-01 12:32:41.149 ERROR [nioEventLoopGroup-1442-1] netty.WebSocketClientHandler - WebSocket Client failed to connect. Invalid handshake response getStatus: 429 Too Many Requests
2018-11-01 12:32:41.152 ERROR [nioEventLoopGroup-1428-1] netty.WebSocketClientHandler - WebSocket Client failed to connect. Invalid handshake response getStatus: 429 Too Many Requests
2018-11-01 12:32:41.153 ERROR [nioEventLoopGroup-1438-1] netty.WebSocketClientHandler - WebSocket Client failed to connect. Invalid handshake response getStatus: 429 Too Many Requests
2018-11-01 12:32:41.154 ERROR [nioEventLoopGroup-1440-1] netty.WebSocketClientHandler - WebSocket Client failed to connect. Invalid handshake response getStatus: 429 Too Many Requests
2018-11-01 12:32:41.155 ERROR [nioEventLoopGroup-1439-1] netty.WebSocketClientHandler - WebSocket Client failed to connect. Invalid handshake response getStatus: 429 Too Many Requests
2018-11-01 12:32:41.155 ERROR [nioEventLoopGroup-1436-1] netty.WebSocketClientHandler - WebSocket Client failed to connect. Invalid handshake response getStatus: 429 Too Many Requests
2018-11-01 12:32:41.164 ERROR [nioEventLoopGroup-1441-1] netty.WebSocketClientHandler - WebSocket Client failed to connect. Invalid handshake response getStatus: 429 Too Many Requests

I have opened a lot of sockets (about 20 socket per each curency pair subscription) and after several network problems with reconections bitfinex blocked my because handshake limit reached.
As you can see there is infinite error loop WITHOUT any delay between reconect actions.
So it would be great to detect exchange blocking error and sleep long time (much more than retryDuration) before reconect. I would think how to improve it.

badgerwithagun:
I've stopped testing for now.  Will wait for you to come up with something, @pchertalev !

kafkapre:
@badgerwithagun @pchertalev  I absolutely disagree to block this pull request because "Too Many Request" problem. This problem can be solved by user of library himself or in another pull request, and memory leak fix is quite urgent, so please move forward this pull request.

Flemingjp:
@kafkapre Some help with this current issue may help move this PR along

badgerwithagun:
I completely agree with @kafkapre  that the issue of rate limiting in general is not the responsibility of this library.

BUT there is a different issue here, which is that automatic reconnection triggered by `NettyStreamingService` can cause rate limit bans without the client code being aware of it.  That does need to be fixed.

What I'm not clear of from @pchertalev's comments is whether this PR makes this problem more likely or not.  If there is no change in the risk as a result of this fix, then yes, let's go ahead and get this in.

pchertalev:
I have tested and commited fix of 'Event loop shut down' (see my last commit). I can't more reproduce error "Event loop shut dow".

pchertalev:
> I completely agree with @kafkapre that the issue of rate limiting in general is not the responsibility of this library.
> 
> BUT there is a different issue here, which is that automatic reconnection triggered by `NettyStreamingService` can cause rate limit bans without the client code being aware of it. That does need to be fixed.
> 
> What I'm not clear of from @pchertalev's comments is whether this PR makes this problem more likely or not. If there is no change in the risk as a result of this fix, then yes, let's go ahead and get this in (and raise an issue to avoid rate limit bans on socket reconnection).

Yes, my PR makes problem more likely, but I have commited fix which allows to avoid it at all.

pchertalev:
Guys, can we approve this pool request?

Flemingjp:
@pchertalev I'd like @badgerwithagun to follow up on your new commits, before we approve this pull request.

badgerwithagun:
And I shall! Should have some time in the next few days.

pchertalev:
> OK, got this into my live environment. Can be pretty sure it's good after a few connections and reconnections over the next 24 hours.

Thanks!

badgerwithagun:
This is stable, and the code changes make sense. This is good for merge, @Flemingjp ,

badgerwithagun:
Thank you for taking the time to help push this through, @pchertalev!

pchertalev:
Thank you guys!

PR #252, opened 2018-11-01
Remove xchange-gdax

BenjaminTrapani:
Remove this exchange because the xchange library recently removed it in favor of coinbase pro (and it was breaking the build locally with the updated xchange library). Would you be interested in a coinbase pro implementation? It looks like they provide a streaming API.


badgerwithagun:
HI @BenjaminTrapani. This has been discussed on #205.

We're planning on deprecating/removing GDAX and replacing with Coinbase Pro, matching what the XChange project has done.  We're a little short on contributors right now so it's not happening as quickly as we'd like.

badgerwithagun:
In the meantime, we can't remove the GDAX implementation as plenty of people are using it and it's the only way to access Coinbase Pro.

badgerwithagun:
Implemented by #261

PR #268, opened 2019-01-13
XChange 4.3.14

badgerwithagun:
Not much in the changes list that's likely to affect xchange-stream. Lots of useful fixes.


Flemingjp:
This prompt a 4.3.14 release here too soon, if appropriate

dozd:
Made the release of xchange-stream right now but as the version of 4.3.15.

4.3.14 failed during the uploading process to the maven central and i was not able to recover it in the staged repository :-(.

PR #271, opened 2019-01-18
Bankera implementation

MrNomNom:


xchange-bankera/src/main/java/info/bitrich/xchangestream/bankera/BankeraStreamingMarketDataService.java
```diff
@@ -0,0 +1,84 @@
...
+  private final BankeraMarketDataService marketDataService;
+
+  public BankeraStreamingMarketDataService(BankeraStreamingService service, BankeraMarketDataService marketDataService) {
+    this.service = service;
+    this.marketDataService = marketDataService;
+  }
+
+  @Override
+  public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
+    BankeraMarket market = getMarketInfo(currencyPair);
```
badgerwithagun: Does Bankera send the full order book with every update? No deltas?

MrNomNom: Yes it does send full orderbook for selected market.

Issue #274, opened 2019-01-27
[WORK IN PROGRESS] Generic authenticated stream API


badgerwithagun:
# Introduction
I intend to create `StreamingTradeService` and `StreamingAccountService` interfaces to accompany `StreamingMarketDataService`, mirroring XChange's core APIs.

These will deliver the following:

* Balance updates
* Authenticated (user) trades
* Order updates/execution reports

This has been a long time coming, but I'm now working on it.  I'll will publish regular PRs to merge work into this repo as I go, status updates here, and fork URLs for anyone that wants to use the work in progress.

Please post here if you have thoughts on the API or implementation approach.

# Approach

PRs #273 (formerly #209), #267, #246 and #244 all have taken different approaches to user-authenticated streams. I am trying to standardise these around the approach of #246 and #273 to follow on from #160, which is:

* [x] If the same socket as the public stream can be "elevated" to an authenticated stream, it should be, rather than opening multiple streams
* [x] xchange-stream will open an authenticated stream (or elevate the public one) if there are API key details provided in the `ExchangeSpecification`
* [x] Supported exchanges will provide a "level 2", exchange-specific API, exposing the raw channel data on public methods.  So, for example, `BinanceStreamingTradeService` will have a `getExecutionReports()` method which returns instances of the execution report data structure in the [binance API](https://github.com/binance-exchange/binance-official-api-docs/blob/master/user-data-stream.md) docs.
* [x] Where possible, the exchanges will support the generic "level 1" APIs detailed below by combining, filtering and transforming the output of their level 2 APIs.
* [x] All data will be provided in a `PublishSubject`-like fashion.  If you're subscribed to the `Observable`, you get the data. If you're not, you don't. Nothing will be buffered.
* [x] I will create new `StreamingTradeService` and `StreamingAccountService` interfaces and move the new methods from `StreamingMarketDataService` to bring the location of these methods in line with XChange.
* [x] The streams contain changes/deltas only
* [x] The streams have no guarantee that updates won't be missed, and no way to tell if they have
Therefore, to use the streams reliably, you should combine these streams with polling to make sure any local snapshots you create are re-synced regularly

# API changes

The following new methods will be created:

| Service | Signature  | Description   | Notes |
|---|---|---|---|
[...]

PR #275, opened 2019-01-28
Add streaming order book support for CEX.IO

BenjaminTrapani:
The code still needs to be formatted per mvn com.coveo:fmt-maven-plugin:format. The original source was not formatted per this extension, so let's revert commit 9eb03e03f2c1cc3df8c33b9ee5e125adfa7bfd1f right before merging to restore the correct format.


xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/dto/CexioWebSocketOrderBookSubscriptionData.java
```diff
@@ -0,0 +1,27 @@
...
+	final boolean subscribe;
+	
+	@JsonProperty
+	final int depth;
+
+	public CexioWebSocketOrderBookSubscriptionData(CurrencyPair currencyPair, boolean subscribe) {
+		pair = new ArrayList<String>(Arrays.asList(currencyPair.base.toString(), 
+				currencyPair.counter.toString()));
+		this.subscribe = subscribe;
+		this.depth = 10; //-1 for full depth
```
BenjaminTrapani: This might be worth exposing at the API layer. Not sure what is the best way to do so because the xchange API does not provide a way of doing this now.

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingMarketDataService.java
```diff
@@ -1,18 +1,74 @@
...
+        	}
+			return retVal;
+		}
+    }
+    
+	@Override
     public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
-        throw new NotYetImplementedForExchangeException();
+		
+		streamingOrderDataService.waitForAuthIfNeeded();
```
badgerwithagun: Why are you authenticating?  Most exchanges don't need it for public data like order books.

Might be worth having a read of #274 to make sure this is heading in the same direction (it looks like it is, at a glance)

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
```diff
@@ -23,44 +30,139 @@
...
         super(apiUrl, Integer.MAX_VALUE);
+        this.isTest = isTest;
     }
-
+    
+    public void waitForAuthIfNeeded() {
+    	authLock.lock();
+    	try
+    	{
+    		while (!isAuthenticated && !isTest)
```
badgerwithagun: Could you remove the `isTest` argument and simply use `apiKey != null`?  That is, authenticate if you have authentication details, and don't authenticate if you don't have them (matches other exchanges - see #274)

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
```diff
@@ -23,44 +30,139 @@
...
+			// TODO Auto-generated catch block
+			e.printStackTrace();
+		}
+    	finally
+    	{
+    		authLock.unlock();
+    	}
+    }
+    
+    public void signalAuthComplete() {
```
badgerwithagun: This lock processing would be much more cleanly implemented with a `CountdownLatch`.

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioAdapters.java
```diff
@@ -71,4 +78,53 @@ private static BigDecimal adaptAmount(BigDecimal amount) {
         }
         return status;
     }
+    
+    private static List<LimitOrder> updateOrInsertQuantityForPrice(List<LimitOrder> existingOrders,
```
badgerwithagun: Apologies if I've not read this in enough detail, but are you aware of all the `update` methods on `OrderBook` which are intended to do this sort of thing?  Is this custom logic definitely necessary?

In some cases it is; some exchanges are weird.  Just want to check.

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioAdapters.java
```diff
@@ -71,4 +78,53 @@ private static BigDecimal adaptAmount(BigDecimal amount) {
...
+    		} 
+    		else 
+    		{
+    			existingOrders.add(insertIndex, orderForUpdatedQuantityPrice);
+    		}
+    	}
+    	return existingOrders;
+    }
+    
+    protected static CurrencyPair adaptCurrencyPairString(String currencyPairString) {
```
badgerwithagun: I'd be very surprised if XChange doesn't already provide a method for this!

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
badgerwithagun: There's another issue here that there is no escape from this lock in the event of a disconnect.  The usual pattern is something like this:

```
while (latch.await(5, SECONDS)) {
  if (disconnected)
    break;
}
```
I wonder, though, if this would be better implemented by performing the auth negotiation _before_ returning complete in the `Completable` returned by `connect()`.  That way you can be sure that this handshaking process will definitely have completed before the caller starts asking for subscriptions.

I suspect a few other exchanges are vulnerable to this issue (particularly Bitfinex, which I'm working on at the moment), so thank you for highlighting it!

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingMarketDataService.java
BenjaminTrapani: For some reason Cex.io requires users to authenticate before gaining access to the order book. This change seems to follow the logic outlined in #274 so far as elevating this stream to an authenticated one, but the mechanism isn't very clean. Since all Cex.IO APIs require authentication, I'll use the connect completable to signal that the authentication has completed.

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
BenjaminTrapani: Done in d1c7ab91daaa8867f279993a72e812c2e3dd4e90

BenjaminTrapani: We do not need an isTest argument for the existing test suite because none of the tests connect to the exchange, so bypass waiting on the returned completable.

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
BenjaminTrapani: The Completable abstract class provides blocking await functions and an overload for a timeout. Callers can wait on the connect for a specific duration before deciding that the connect failed. We could also trigger an error on the completable on auth failures, which might be better. Should I add code to invoke OnError on auth failures?

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioAdapters.java
BenjaminTrapani: OrderBook update did exactly what I needed. This is fixed in d1c7ab91daaa8867f279993a72e812c2e3dd4e90

xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioAdapters.java
BenjaminTrapani: They do but it is private :( I copied their implementation exactly with a note in the code about the duplication. I can submit a PR to them to make the function protected if that would be useful.

BenjaminTrapani:
@badgerwithagun These changes should be ready to go now. I reformatted xchange-cexio per the mvn format used in xchange in the final commit, so it is easiest to read the updates by looking at the diffs for the individual commits before 3e16614.

BenjaminTrapani:
@badgerwithagun Ping on this one. Is there anything I missed in the feedback? How does the most recent iteration look? If it looks alright, I'll sync up with latest develop.

badgerwithagun:
Hey @BenjaminTrapani, sorry, I've been busy. Will look at this ASAP.

badgerwithagun:
OK, sorry about the delay  @BenjaminTrapani . It's clear there's no-one else using CEX.IO, so I'm just going to take a leaf out of the XChange project's book and merge this.  It can't hurt if only you are affected!

Issue #278, opened 2019-02-01
Number of nioEventLoopGroup threads keeps on increasing if connection is not succesful


KeshavSirohi:
Faced this issue when GDAX kept on trying for re-connection. Older nio threads were not getting shutdown. Thread dump showed more than 100 nioEventLoopGroup thread in Runnable state.
Below are the websocket logs:
2019-01-31 01:00:40,131  INFO nioEventLoopGroup-138-1 WebSocket Client connected!
2019-01-31 01:00:40,131  INFO nioEventLoopGroup-137-1 WebSocket Client connected!
2019-01-31 01:00:40,131  WARN nioEventLoopGroup-137-1 Resubscribing channels
2019-01-31 01:00:40,131  WARN nioEventLoopGroup-138-1 Resubscribing channels
2019-01-31 01:01:18,342  INFO nioEventLoopGroup-135-1 WebSocket Client disconnected!
2019-01-31 01:01:18,342  INFO nioEventLoopGroup-135-1 Reopening websocket because it was closed by the host
2019-01-31 01:01:18,342  INFO nioEventLoopGroup-138-1 WebSocket Client disconnected!
2019-01-31 01:01:18,342  INFO nioEventLoopGroup-138-1 Reopening websocket because it was closed by the host
2019-01-31 01:01:20,545  INFO globalEventExecutor-1-162 Connecting to wss://ws-feed.gdax.com:-1
2019-01-31 01:01:20,547  INFO globalEventExecutor-1-162 Registering GDAXWebSocketClientHandler
2019-01-31 01:01:20,548  INFO globalEventExecutor-1-162 Connecting to wss://ws-feed.gdax.com:-1
2019-01-31 01:01:20,550  INFO globalEventExecutor-1-162 Registering GDAXWebSocketClientHandler
2019-01-31 01:01:21,292  INFO nioEventLoopGroup-139-1 WebSocket Client connected!
2019-01-31 01:01:21,292  WARN nioEventLoopGroup-139-1 Resubscribing channels
2019-01-31 01:01:21,297  INFO nioEventLoopGroup-140-1 WebSocket Client connected!
2019-01-31 01:01:21,297  WARN nioEventLoopGroup-140-1 Resubscribing channels
2019-01-31 01:01:40,083  INFO nioEventLoopGroup-137-1 WebSocket Client disconnected!
2019-01-31 01:01:40,083  INFO nioEventLoopGroup-137-1 Reopening websocket because it was closed by the host
[...]

Issue #276, opened 2019-01-29
Minor synchronization issues with ConcurrentHashMap


berserkk:
`ConcurrentHashMap` isn't fully correctly used in the code.
Yeah, you will not get `ConcurrentModificationException`, but there's possibility to get other issues like lost updates or `NullPointerException`. Of course, possibility is very low, but non-zero.

Here's examples:

https://github.com/bitrich-info/xchange-stream/blob/3819363226d16b193fe8c91ced11f958921ec1f7/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java#L309-L311

`containsKey` and `put` are 2 different atomic operations. But the whole operation should be executed atomically. You can use [putIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#putIfAbsent-K-V-) or [computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-) methods to fix it.

https://github.com/bitrich-info/xchange-stream/blob/3819363226d16b193fe8c91ced11f958921ec1f7/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java#L319-L322

Can be replaced with something like:

```
if (channels.remove(channelId) != null) {
    sendMessage(getUnsubscribeMessage(channelId)); 
}
```

but `getUnsubscribeMessage` must not use `channels` for fetching value by removed `channelId`.

https://github.com/bitrich-info/xchange-stream/blob/3819363226d16b193fe8c91ced11f958921ec1f7/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java#L327-L329

Just iterate over [entrySet](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#entrySet--).

Should also check `ConcurrentHashMap` usages in other places.

badgerwithagun:
Fixing this now

badgerwithagun:
@berserkk, would you be interested in providing your review on #282?

berserkk:
@badgerwithagun, sure, I'll check.

PR #282, opened 2019-02-05
Netty concurrency fixes, more consistent thread shutdown on errors

badgerwithagun:
Address #276, #278.

I've not been able to verify that this definitely prevents the thread leaks described in #278 since I've been unable to replicate them, but I've traced through and identified these additional places that a leak is possible,

Regarding #276, this is pretty much as described on the issue.

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -232,10 +232,10 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
                 if(f.isSuccess()) {
                     isManualDisconnect = false;
                 }
-                shutdownEventLoopGroup(completable);
+                shutdownEventLoopGroup(() -> completable.onError(t));
             });
         } else {
-            shutdownEventLoopGroup(completable);
+            shutdownEventLoopGroup(() -> completable.onError(t));
         }
     }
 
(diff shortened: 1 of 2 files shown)


service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -312,28 +328,27 @@ public void sendMessage(String message) {
             if (webSocketChannel == null || !webSocketChannel.isOpen()) {
                 e.onError(new NotConnectedException());
             }
-
-            if (!channels.containsKey(channelId)) {
+            channels.computeIfAbsent(channelId, chanId -> {
```
berserkk: Looks like `sendMessage` is non-blocking, but anyway, I think, this computation can be safely moved outside of lambda. So, it can be rewritten like this:

```
Subscription newSubscription = new Subscription(e, channelName, args);
if (channels.putIfAbsent(channelId, newSubscription) == null) {
    try {
        sendMessage(getSubscribeMessage(channelName, args));
    } catch (IOException throwable) {
        e.onError(throwable);
    }
}
```

The reason for this is that underlying lock will be taken for shorter period of time.

badgerwithagun:
Yeah, I tried to keep the changes to a minimum bit the more I look at this code the more I worry about it. It's too easy to introduce race conditions and all the null checks everywhere just indicate a deeper problem. Too much independently mutable state.

I'll have a bit more of a play.

badgerwithagun:
I'm made a lot more progress with this as part of #274, since that needs more substantial changes in this area. I'm going to close this PR without merge.

Issue #284, opened 2019-02-19
Bitfinex corrupted orderbooks


davidjirovec:
I am using most recent version from develop, today I've again ran into issues with corrupted orderbooks mentioned in #269. I can see clearly in my logs that top value in orderbook was stuck there for several hours, while orders were constantly changing.

I've tried to enable websocket sequencing, result is here: https://github.com/davidjirovec/xchange-stream/tree/sequencing-test - just run BitfinexManualExample.

Logs like this start appearing:

```
[153,[4014,2,0.28261943],593]
[153,[4008.3,2,2.40802452],594]
[153,[4008.2,2,2.3670261],595]
[153,[4054.4,2,-2.52],596]
[153,[4033.3,0,-1],597]
[153,[4033.2,1,-0.49633561],598]
[153,[4029.8,2,-1.72184484],599]
[153,[4008.2,0,1],600]
[153,[4025.8,1,0.3866601],601]
[153,[4025.6,1,0.2],602]
[153,[4024.8,0,1],603]
[153,[4008.2,2,2.3670261],604]
[153,[4029.8,3,-2.22184484],606]
Wrong sequence 606 vs 605
[153,[4047.9,2,-0.62546047],607]
Wrong sequence 607 vs 606
[153,[4008.2,0,1],608]
Wrong sequence 608 vs 607
[153,[4024.8,1,0.39387914],609]
Wrong sequence 609 vs 608
```

It seems message 605 got lost somehow.

Maybe I am doing something wrong, but compared to official nodejs sample https://github.com/bitfinexcom/bitfinex-api-node/blob/bf00a353d90578797a23fac996eed5e9fee81b91/examples/ws2/sequencing.js I don't see anything important.

Some time ago I've also implemented orderbook checksums, these sometimes fail too, but it usually takes days to happen and I prefer not to use it to make my bot fast as possible.

Any ideas?

badgerwithagun:
Most exchanges don't guarantee that there won't be gaps in websocket feeds (binance certainly doesn't and neither does Coinbase Pro). The binance order book implementation does what they (and CBP) recommend, which is to detect gaps and rebase from the Rest API. Not sure if this definitely applies to Bitfinex's order books.

fwiw, I am planning to implement guaranteed safety on the new authenticated channels (see my open issue) by applying whatever exchange specific processing is required to do so. This means making rest calls, which means we will want another ExchangeSpecification parameter to allow rate limiting of these, as we do for socket reconnections.

[...]

Issue #183, opened 2018-05-14
[BitMEX] orderBookL2 feed bugs out at 12:00 UTC


somguyth:
Has anyone been having problems with book building off BitMEX's orderBookL2 feed recently?

Just after 12:00 UTC (which is the funding time) on 9th May, 11th May and 14th May (today), the exchange started sending book updates that refer to non-existent order/level ids. 

For example, just 2 hours ago (14th May at 12:00:40 UTC) I got a book update for id 8791580900 that tried to amend the size of the order. However, the last message received with this specific id was a **delete** message at 11:24:24 UTC, so the fact the exchange is sending an amend message after a delete doesn't make sense.

jkolobok:
@somguyth have you tried comparing it with what REST api returns?

Flemingjp:
This bug is several months old - needs testing and validating to see if it still bug. It is likely it was a server side bug that may have been fixed by now.

Issue #137, opened 2018-03-17
OkCoinStreamingExchange throws java.lang.NullPointerException


quorak:
```
2018-03-17 10:49:48.548  INFO 56214 --- [           main] i.b.x.okcoin.OkCoinStreamingExchange     : Calling Remote Init...
2018-03-17 10:49:48.548  INFO 56214 --- [           main] i.b.x.okcoin.OkCoinStreamingExchange     : No remote initialization implemented for OKCoin. The exchange meta data for this exchange is loaded from a json file containing hard-coded exchange meta-data. This may or may not be OK for you, and you should understand exactly how this works. Each exchange can either 1) rely on the hard-coded json file that comes packaged with XChange's jar, 2) provide your own override json file, 3) properly implement the `remoteInit()` method for the exchange (please submit a pull request so the whole community can benefit) or 4) a combination of hard-coded JSON and remote API calls. For more info see: https://github.com/timmolter/XChange/wiki/Design-Notes#exchange-metadata
2018-03-17 10:49:48.561  INFO 56214 --- [           main] i.b.x.okcoin.OkCoinStreamingService      : Connecting to wss://real.okcoin.com:10440/websocket
2018-03-17 10:49:50.453  INFO 56214 --- [ntLoopGroup-2-1] i.b.x.s.netty.WebSocketClientHandler     : WebSocket Client connected!
2018-03-17 10:49:50.455  INFO 56214 --- [           main] i.b.x.okcoin.OkCoinStreamingService      : Subscribing to channel ok_sub_spot_btc_usd_depth
java.lang.NullPointerException
	at info.bitrich.xchangestream.service.netty.NettyStreamingService.handleChannelMessage(NettyStreamingService.java:276)
	at info.bitrich.xchangestream.service.netty.NettyStreamingService.handleMessage(NettyStreamingService.java:266)
	at info.bitrich.xchangestream.okcoin.OkCoinStreamingService.handleMessage(OkCoinStreamingService.java:47)
	at info.bitrich.xchangestream.okcoin.OkCoinStreamingService.handleMessage(OkCoinStreamingService.java:11)
	at info.bitrich.xchangestream.service.netty.JsonNettyStreamingService.messageHandler(JsonNettyStreamingService.java:39)
	at info.bitrich.xchangestream.service.netty.NettyStreamingService$$Lambda$128/1729958231.onMessage(Unknown Source)
[...]

PR #360, opened 2019-07-23
Added fragmented text message handling in WebSocketClientHandler

Nephtys:
When the received message was fragmented, the first part was sent to the message handler and the rest of the message was ignored. 
The first message has type TextWebSocketFrame and the following have type ContinuationWebSocketFrame. Each frame has a boolean finalFragment to indicate if the message was fully received.
See RFC 6455 section 5.4 and 5.2, and Netty WebSocket08FrameDecoder for more information.

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/WebSocketClientHandler.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/WebSocketClientHandler.java
@@ -1,8 +1,5 @@
 package info.bitrich.xchangestream.service.netty;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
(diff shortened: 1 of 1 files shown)


Nephtys:
Issues mentioning intermittent failures or lost messages may be related

badgerwithagun:
@Nephtys: The change looks sensible. Did you encounter this issue with a specific exchange, or is this just precautionary?

Nephtys:
We encountered this issue with snapshot messages on the LGO exchange websocket, but it's not an exchange specific case, it's how the websocket protocol was designed and any exchange using message framing will encounter this issue.
The resulting errors can be silent because the message is received incomplete, so it can be ignored by the following handlers (in fact, in case of a JSON payload, the JSON will be incomplete, resulting in an exception in the JSON parser and the message never sent to the exchange specific handler (or null). That's why that could be the reason for some other open bugs like "message lost" problems in xchange-stream.
See RFC 6455 section 5.4 and 5.2, and Netty WebSocket08FrameDecoder for more information on how the websocket protocol is built.

badgerwithagun:
All understood; I did read up on this part of the protocol before asking. I'm just wondering which issues this fix might close, since the original reporters might find it interesting. #284 perhaps.  I found it odd that I've never seen it, but I agree; it's probably that it's manifesting in lots of different ways.

In any case, this looks sensible, as I said, so LGTM.

davidjirovec:
Yes, it looks it could help. I am still running my custom xchange-stream bitfinex version with checksums. Still checksum error a few times a day. I'll try changes from this PR, after a few days I can say if it helped.

badgerwithagun:
Thanks @Nephtys and @davidjirovec

Nephtys:
Yes you're right, maybe https://github.com/bitrich-info/xchange-stream/issues/284 and maybe https://github.com/bitrich-info/xchange-stream/issues/183 and https://github.com/bitrich-info/xchange-stream/issues/137

Thank you for your review !

davidjirovec:
Testing, checksum errors still sometimes show up :-|

PR #371, opened 2019-08-06
[Bankera] upgrade the parent version to fix the build

garciapd:
The parent version was set to an old one and the build was broken


garciapd:
Thanks @badgerwithagun

Issue #361, opened 2019-07-24
Kraken Websocket orderbook is inconsistent [Bug!]


makarid:
Hello, i was testing the Kraken orderbook websocket and i found that the prices and volume is inconsistent with the real values. I can send you printscreen if you like. Does anyone found that also? I am waiting for the response of the @pchertalev .Thanks

makarid:
I just found the issues.

1) There is an issue when the last digit of the price before dot is zero. For example when the price of bitcoin is 8750.0000, the method stripTrailingZeros() of BigDecimal gives as output 875+E3. I have fixed this by deleting the stripTrailingZeros() method.

2) When the Kraken book subscription sends a message in order to update values in both bid and ask orderbook, the current implementation updates only the one side of the book( the first one, so only ASK). I have fix this issue and i will make a PR shortly.

badgerwithagun:
Thanks @makarid. I'll be happy to merge the PR if one of our other Kraken users can confirm it works.

pchertalev:
> I just found the issues.
> 
> 1. There is an issue when the last digit of the price before dot is zero. For example when the price of bitcoin is 8750.0000, the method stripTrailingZeros() of BigDecimal gives as output 875+E3. I have fixed this by deleting the stripTrailingZeros() method.

It not correct! 8750.0000 with  stripTrailingZeros returns 8.75E+3  (not 875+E3)!  Where is the bug here? 8750.0000 fully equals 8.75E+3.

pchertalev:
> I just found the issues.
> 2. When the Kraken book subscription sends a message in order to update values in both bid and ask orderbook, the current implementation updates only the one side of the book( the first one, so only ASK). I have fix this issue and i will make a PR shortly.

I have long testing WS Kraken Order book in comparison vs rest order book and it works fully correct in asks and bids.  There is only one issue I found - Kraken rest order book use rounding order book up to 3 decimal points but WS order book not - WS order book is more accurate

pchertalev:
@makarid Your pool request is not readable - it contains a lot of durty commits from not merged branches. Could you create new pure fork from xchainge-streaming project, commit your changes and create new pool request?

makarid:
@pchertalev thanks for your response. 

About the first issue, maybe it is my bad, but i have made this test and it fails:
        BigDecimal price = new BigDecimal(8500).stripTrailingZeros();
        System.out.println(price);
[...]

PR #372, opened 2019-08-07
#361 Kraken WebSocket Implementation Bug: sometimes incremental updat…

pchertalev:
…es are skipped


pchertalev:
@badgerwithagun Graham, It is very important PR, because sometimes Kraken orderbook is loosing updates. I repoduced this bug and fixed in my fork. We defenetly must this to merge.

PR #385, opened 2019-08-19
[LGO] Add LGO exchange websocket integration

Nephtys:
Hi, 

Please add support for LGO websocket API.
https://lgo.group/

LGO XChange support is available since release 4.3.21.
Thank you.


badgerwithagun:
Sorry about the delay reviewing @Nephtys . I was halfway through it last week and got dragged onto something else. I'll try and finish today.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingService.java
```diff
@@ -0,0 +1,52 @@
...
+    protected DefaultHttpHeaders getCustomHeaders() {
+        DefaultHttpHeaders headers = super.getCustomHeaders();
+        String timestamp = String.valueOf(System.currentTimeMillis());
+        headers.add("X-LGO-DATE", timestamp);
+        String auth = signatureService.digestHeader(this.apiUrl + "/", timestamp);
+        headers.add("Authorization", auth);
+        return headers;
+    }
+
+    private final LgoSignatureService signatureService;
```
badgerwithagun: Could these be moved to the top?

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
```diff
@@ -0,0 +1,287 @@
...
+    public Observable<Order> getOrderChanges(CurrencyPair currencyPair, Object... args) {
+        return getOrderBatchChanges(currencyPair)
+                .flatMap(Observable::fromIterable);
+    }
+
+    /**
+     * Get an up-to-date view of all your open orders after each LGO batch execution.
+     * First sending will be the actual open orders list.
+     */
+    public Observable<OpenOrders> getOpenOrders(CurrencyPair currencyPair) {
```
badgerwithagun: In my experience, this is very unlikely to be stable.  There's no way for Netty to tell if a web socket has dropped and you have missed messages, unless the exchange provides a sequentially numbered heartbeat and  a clear guide about how to refresh from the REST API in such a way that you can safely discard websocket messages in the meantime. Binance has an excellent guide for how to do this for their order book API, but order books are updated constantly so there are plenty of sync opportunities. Open orders don't update very often (unless you're doing arbitrage).

I always write code that _both_ polls the REST API (on a longer timescale) _and_ monitors the websocket (to get faster updates) and manages the corresponding snapshot in a layer above both xchange-stream and XChange.  The regular polling ensures that if inconsistencies appear, they don't last long in memory.

This approach looks like it has no self-correction mechanism (although I can't be sure without having read more about LGO, so please correct me if I am wrong) in which case if this cache gets out of date, it will stay out of date until the open orders are changed while the web socket is properly connected.

When discussing this API, we spent quite a long time trying to decide whether xchange-stream should maintain a snapshot of the orders like this, make REST API calls for synchronisation and so on, but couldn't find any way of doing it reliably across exchanges so we avoided it.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
```diff
@@ -0,0 +1,287 @@
...
+        return getBatchOrderEvents(currencyPair)
+                .filter(lgoOrderEvent -> lgoOrderEvent.getType().equals("match"))
+                .map(matchEvent -> LgoAdapter.adaptUserTrade(currencyPair, (LgoMatchOrderEvent) matchEvent));
+    }
+
+    /**
+     * Receive all events for the selected currency pairs. Merges batch order events and ack (AFR)
+     * events.
+     */
+    public Observable<LgoOrderEvent> getAllOrderEvents(Collection<CurrencyPair> currencyPairs) {
```
badgerwithagun: Rename to `getRawAllOrderEvents` in line with convention

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
```diff
@@ -0,0 +1,287 @@
...
+                .reduce(Observable::mergeWith);
+        return batchObservable.isPresent() ? ackObservable.mergeWith(batchObservable.get()) : ackObservable;
+    }
+
+    /**
+     * Get ack for your placed orders. "received" events indicate the orderId associated to your
+     * order, if you set a reference on order placement you will have it in the event. "failed" events
+     * indicate that the order could not be read or was invalid and not added to a batch.
+     */
+    public Observable<LgoOrderEvent> getReceivedOrderEvents() {
```
badgerwithagun: Rename to `getRawReceivedOrderEvents` in line with convention

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * Get all events of your orders happened during batch execution, for a currency pair. "pending"
+     * events indicate that the order was added to a batch and received by the execution engine.
+     * "invalid" events indicate that the order was not suitable for execution. "match" events
+     * indicate that the order did match against another order. "open" events indicate that the order
+     * entered the order book. "done" events indicate that the order was filled, canceled or
+     * rejected.
+     */
+    public Observable<LgoOrderEvent> getBatchOrderEvents(CurrencyPair currencyPair) {
```
badgerwithagun: Rename to `getRawBatchOrderEvents` in line with convention

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
```diff
@@ -0,0 +1,287 @@
...
+     * @return true
+     */
+    public boolean cancelOrder(String orderId) throws IOException {
+        Long ref = nonceFactory.createValue();
+        LgoPlaceCancelOrder lgoOrder = new LgoPlaceCancelOrder(ref, orderId, new Date().toInstant());
+        placeOrder(ref, lgoOrder);
+        return true;
+    }
+
+    private String placeOrder(Long ref, LgoPlaceOrder lgoOrder) throws JsonProcessingException {
```
badgerwithagun: This is interesting! Not seen exchanges which permit placing orders over the websocket before.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
BodySplash: Interesting. 
LGO sends monotonic batch_id on each message, so a missed message could be detected. In that case, it's just a matter of unsubscribing/subscribing to this channel to receive a fresh snapshot. We could implement that. What do you think ? 
In certain conditions, LGO will also send a snapshot if it thinks some data may have been missed. That case is already dealt with.

badgerwithagun: That does sound good, as long as the {{StreamingTradeService}} manages the disconnection/reconnection transparently, _without disrupting the existing `Observable`s_ (so the client isn't required to close everything and fetch new `Observable`s).  That would mean switching _all_ the exposed `Observable`s on all of the `Streaming...Service` implementations to `Subject`s.  That is a big change!

However, it still doesn't solve the whole problem. You've still got the situation where:

# Socket opens
# Initial snapshot is sent
# Websocket (silently) lost
# Update is sent
# Websocket (silently) recovers
# Time passes where no order updates are sent and the open orders are out of date
# Update is sent - missing message is detected and recovery occirs

What I found was that this happens surprisingly often, and given that open orders don't necessarily update very often, that time gap can be very large.

badgerwithagun: Basically I've got no issue with whatever approach you want to use for this - my only concern is that if the feature appears in this library, it needs to be solid and work without the client having to do anything "special".  If we can come up with such a solution, fantastic. If not, it might be safer to omit it for the time being.

BodySplash: I see you point, thank you for your feedback. We'll look at our options in the next few days. 

Thinking out loud : A traditional way in socket land to deal with connection loss detection is to send bi-directional keep-alive messages. I guess one could use the standard PING mechanism of websockets to implement that, to make sure that in the absence of messages after a given time, it's not due to connectivity issues.

badgerwithagun: Yes indeed. In fact, IIRC, there's support in NettyStreamingService already from when I tried to do exactly this (and gave up!)

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoAdapter.java
```diff
@@ -0,0 +1,166 @@
...
+            Currency currency = Currency.getInstance(balance.get(0));
+            BigDecimal available = new BigDecimal(balance.get(1));
+            BigDecimal escrow = new BigDecimal(balance.get(2));
+            BigDecimal total = available.add(escrow);
+            return new Balance(currency, total, available, escrow);
+        }).collect(toList());
+    }
+
+    private static Order.OrderType parseTradeType(LgoTrade lgoTrade) {
+        //the XChange API requires the taker order type whereas the LGO API gives the maker order type
```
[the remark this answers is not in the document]
Nephtys: indeed!

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoAdapter.java
```diff
@@ -0,0 +1,166 @@
...
+        return new MarketOrder.Builder(parseOrderType(orderEvent), currencyPair)
+                .id(orderEvent.getOrderId())
+                .originalAmount(orderEvent.getQuantity())
+                .cumulativeAmount(BigDecimal.ZERO)
+                .orderStatus(Order.OrderStatus.PENDING_NEW)
+                .timestamp(orderEvent.getTime())
+                .build();
+    }
+
+    static List<LgoOrderEvent> adaptOrderEvent(List<LgoUserUpdateData> data, Long batchId, List<Order> updatedOrders) {
```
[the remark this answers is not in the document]
Nephtys: https://github.com/bitrich-info/xchange-stream/pull/385/commits/7059acb767945daf35fce743ddc097e0b350438e

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoAdapter.java
```diff
@@ -0,0 +1,166 @@
...
+                    break;
+                case "done":
+                    events.add(new LgoDoneOrderEvent(batchId, orderEvent.getType(), orderEvent.getOrderId(), orderEvent.getTime(), orderEvent.getReason(), orderEvent.getCanceled()));
+                    break;
+            }
+        }
+        return events;
+    }
+
+    static LgoOrderEvent adaptReceivedEvent(LgoAckUpdateData data) {
```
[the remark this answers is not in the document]
Nephtys: https://github.com/bitrich-info/xchange-stream/pull/385/commits/7059acb767945daf35fce743ddc097e0b350438e

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
Nephtys: https://github.com/bitrich-info/xchange-stream/pull/385/commits/bf18a67a4e7b047677722216dbbf794927610689

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
Nephtys: https://github.com/bitrich-info/xchange-stream/pull/385/commits/bf18a67a4e7b047677722216dbbf794927610689

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
Nephtys: https://github.com/bitrich-info/xchange-stream/pull/385/commits/bf18a67a4e7b047677722216dbbf794927610689

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingService.java
Nephtys: https://github.com/bitrich-info/xchange-stream/pull/385/commits/bf18a67a4e7b047677722216dbbf794927610689

badgerwithagun:
Hey @BodySplash and @Nephtys.

What are your thoughts on our prior discussion?  Would be nice to get this merged.

arass:
> standard PING mechanism of websockets to implement that

Bunch of WS implementations in Java don't support those. I would recommend using explicit heartbeat messages, as suggested above. 

Great to see LGO integration being added. 
Can you guys merge it ASAP and they fix it in place? 
Future patches would be smaller and easier to deal with. 
And others would be able to contribute as well then.  Right now the rest of us have to wait for this merge to touch it. 

Thanks.

BodySplash:
Sorry @badgerwithagun , a bunch of other priorities has prevented us to dig this issue. We'll try to resume work on that soon.

badgerwithagun:
Thanks @BodySplash

BodySplash:
Hello again. I've been working on some workarounds, not perfect yet. I tried for now just to implement what we talked about only for LGO, because I didn't feel confident enough yet to do some significant changes in xchange stream core. 

Short version : 
* on idle, send a PING message. I simulated some connectivity issue, and XChange-Stream triggered properly a reconnect / resubscribe process, without disposing observables. Still, I would rather have an explicit PONG timeout mechanism. At least, this mechanism check on regular basis if xchange-stream is still able to send message to LGO.
* on out of sync seq/batch_id, resubscribe. After some experiments, I decided to simply send standard subscribe/unsubscribe messages, since LGO will send back full snapshot in that case. It's not ideal, but it's a start. 
Happy to ear your feedback @badgerwithagun

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingService.java
```diff
@@ -0,0 +1,60 @@
...
+    private final String apiUrl;
+
+    LgoStreamingService(LgoSignatureService signatureService, String apiUrl) {
+        super(apiUrl, Integer.MAX_VALUE, Duration.ofSeconds(10), Duration.ofSeconds(15), 1);
+        this.apiUrl = apiUrl;
+        this.signatureService = signatureService;
+    }
+
+    @Override
+    protected void handleIdle(ChannelHandlerContext ctx) {
```
badgerwithagun: Heh, it's a blunt instrument, but probably effective enough!

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
badgerwithagun: OK it looks like you're getting full snapshots every time and aggressively keeping the socket alive, so this is probably going to work.  Fingers crossed! 👍

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Get an up-to-date view of all your open orders after each LGO batch execution.
+     * First sending will be the actual open orders list.
+     */
+    public Observable<OpenOrders> getOpenOrders(CurrencyPair currencyPair) {
+        return getOrderUpdates(currencyPair)
+                .map(u -> u.getAllOpenOrders().values().stream()
+                        .filter(order -> order instanceof LimitOrder)
```
badgerwithagun: Why the filter?  You can pass limit orders to the first parameter of `OpenOrders`' constructor, and stop orders to the second.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoUserBatchSubscription.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private Observable<LgoGroupedUserUpdate> createSubscription() {
+        final ObjectMapper mapper = StreamingObjectMapperHelper.getObjectMapper();
+        return streamingService
+                .subscribeChannel(LgoAdapter.channelName("user", currencyPair))
+                .map(s -> mapper.readValue(s.toString(), LgoUserMessage.class))
+                .scan(new LgoGroupedUserUpdate(), (acc, s) -> {
+                    List<LgoBatchOrderEvent> events = new ArrayList<>();
+                    if (s.getType().equals("update")) {
+                        if (s.getBatchId() != acc.getBatchId() + 1) {
```
badgerwithagun: Same thoughts around this one as the others.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingAccountService.java
```diff
@@ -2,49 +2,73 @@
...
-                    LgoAdapter.adaptBalances(s.getData()).forEach(balance -> wallet.put(balance.getCurrency(), balance));
-                    Collection<Balance> values = wallet.values();
-                    return Wallet.Builder.from(values)
-                            .build();
-                });
+                    if (acc.getSeq() + 1 != s.getSeq()) {
+                        LOGGER.warn("Wrong seq expected {} get {}", acc.getSeq() + 1, s.getSeq());
+                        resubscribe();
+                    }
+                    return acc.applyUpdate(s.getSeq(), updatedBalances);
```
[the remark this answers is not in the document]
BodySplash: This update only sends informations for currencies actually changed. Hopefully, a future update will set things straight, but in the mean time, you are not 100% sure what your balances are. 
On the other hand, for now, LGO only manages a handful of currencies, so it shouldn't happen.

Websockets are TCP based, so, in order delivery is a given. Have you ever observe the behaviour you are describing ?

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingAccountService.java
```diff
@@ -2,49 +2,73 @@
...
+import java.util.List;
 
 public class LgoStreamingAccountService implements StreamingAccountService {
 
+    private static final String CHANNEL_NAME = "balance";
     private final LgoStreamingService service;
-    private final Map<Currency, Balance> wallet = new HashMap<>();
-    private Observable<Wallet> walletObservable = null;
+    private static final Logger LOGGER = LoggerFactory.getLogger(LgoStreamingAccountService.class);
+    private volatile Observable<LgoGroupedBalanceUpdate> subscription = null;
```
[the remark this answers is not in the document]
BodySplash: Sorry, it was in my todo, forgot about that one. 
compareAndSet doesn't accept a provider to lazy create the instance, so synchronized it will be.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingService.java
BodySplash: Indeed :) As a matter of fact, it serves exactly this purpose of sending PING messages everytime we don't have data out of the socket after a given time. 
Of course, without a PONG timeout, it's not perfect, but at least in case on connectivity issues, I noticed that netty does a proper job of  closing everything.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoUserBatchSubscription.java
BodySplash: Which one exactly ?

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
BodySplash: It was more to deal with MarketOrder here, since we don't have Stop orders, and to my knowledge, OpenOrders doesn't support these.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoLevel2BatchSubscription.java
```diff
@@ -0,0 +1,67 @@
...
+                .scan(new LgoGroupedLevel2Update(), (acc, s) -> {
+                    if (s.getType().equals("snapshot")) {
+                        acc.applySnapshot(s.getBatchId(), s.getData());
+                        return acc;
+                    }
+                    if (acc.getLastBatchId() + 1 != s.getBatchId()) {
+                        LOGGER.warn("Wrong batch id. Expected {} got {}.", acc.getLastBatchId() + 1, s.getBatchId());
+                        resubscribe();
+                    }
+                    acc.applyUpdate(s.getBatchId(), s.getData());
```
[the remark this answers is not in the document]
BodySplash: Correct. I couldn't decide what I wanted to do here. 
On the other hand, what is the best way to avoid emitting this one, without disposing the observable ?

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoLevel2BatchSubscription.java
```diff
@@ -0,0 +1,67 @@
...
+                    return acc;
+                })
+                .skip(1)
+                .map(acc -> LgoAdapter.adaptOrderBook(acc.getBidSide(), acc.getAskSide(), currencyPair))
```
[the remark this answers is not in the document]
BodySplash: Good idea, I'll try to work on that.

mdvx:
I will be happy to test this as I have LGO sandbox account - does this require interaction with Ledger Nano? (as per previous APIs)

BodySplash:
thanks @mdvx 
The api doesn't require the ledger, just a secret key and a token. However you would require the ledger at least once to run the lgo desktop app to register your api secret, and then you're all set. 
Happy to help you on those steps if you need it.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoLevel2BatchSubscription.java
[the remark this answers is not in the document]
badgerwithagun: How about wrapping the orderbook on the stream with a simple object containing the orderbook and a `dirty` flag. This flag gets set to `true` in the `scan` if you get an out of sequence update, and can therefore be `filter`ed immediately afterwards and then `map`ped to the order book.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingAccountService.java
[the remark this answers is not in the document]
badgerwithagun: Not with LGO. My only experience is with Binance and Bitfinex order books (where yes, this sort of thing matters).

What I mean, really, is that I can't see the reason for resubscribing.  You're clearly still getting messages. You might have missed one (which is fine, we don't care) or got one out of order (which is is easy to detect and ignore as discussed above).

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingTradeService.java
badgerwithagun: Ah, they return `MarketOrder`s? Fine then.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoUserBatchSubscription.java
badgerwithagun: I can't see the reason for resubscribing. You're clearly still getting messages. You might have missed one (which is fine, we don't care) or got one out of order (which is is easy to detect and ignore as discussed above).

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoLevel2BatchSubscription.java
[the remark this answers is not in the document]
BodySplash: done, as long as filtering out of sync order books updates.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoStreamingAccountService.java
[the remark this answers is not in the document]
BodySplash: removed

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoUserBatchSubscription.java
BodySplash: removed

mdvx: @badgerwithagun Just a note here, in offline conversations with @BodySplash, I found that some messages were being dropped.  I have added this back in my local version, no recommendation yet, since i have not re-reproduced.

badgerwithagun:
@mdvx and @BodySplash - I'm aware that this one has been held up in review for a long time. What's the status?  Is it working?  If so, I'm probably happy to get it merged and we can fix onward from there.

BodySplash:
@badgerwithagun it's working yes. They are a few enhancements that I would like to make, but I think it can wait another PR, and let this initial one be merged. Thanks for all your comments and feedbacks.

xchange-lgo/src/main/java/info/bitrich/xchangestream/lgo/LgoUserBatchSubscription.java
badgerwithagun: Leaving this conversation open pending another PR with fixes.

badgerwithagun:
Marvellous. We got there!

Look forward to follow-ups when it's had a chance to settle in.

PR #395, opened 2019-08-30
Update to latest Xchange version 4.3.22-SNAPSHOT

makarid:


makarid:
I will keep the PRs very simple and easy in order to be merged fast. I have a lot of code waiting to be merged on the side lines. Thanks

badgerwithagun:
DO you need Xchange 22-SNAPSHOT, @makarid, or can you just use Xchange 21?

Moving to 22-SNAPSHOT means we can't do a release until switching to 22.

makarid:
Sorry for my newbie question but can you please tell me which is the main problem that we can not do a release until switching to 22? I am always using the latest of Xchange.

badgerwithagun:
Maven Central does not allow you to publish a release that depends on SNAPSHOT releases of other libraries. You have to depend only on release versions. So, if we switch xchange-stream to an XChange SNAPSHOT release, we will then need to wait for the corresponding final release to come out and switch to that before we can do a release.

Generally it's easier just to avoid SNAPSHOT dependencies altogether unless you are preparing changes which are dependent on changes in the forthcoming release and are happy to wait for that release to come out before releasing yourself.

Hence the question: do you actually need 22 or will 21 do?

makarid:
Thank you so much for your answer. I didn't know that about Maven. I will revert to 21 then.

PR #387, opened 2019-08-20
#386 Bitfinex and Kraken concurrent modification - fix

pchertalev:
#386 fix

PR #396, opened 2019-08-31
[KrakenStreaming] Adding verifyOrderbook method

makarid:
This method is helpful when a spike or extreme price action occurs.

### DIFF
--- a/xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenOrderBookUtils.java
+++ b/xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenOrderBookUtils.java
@@ -1,20 +1,15 @@
 package info.bitrich.xchangestream.kraken;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
 import info.bitrich.xchangestream.kraken.dto.KrakenOrderBook;
 import info.bitrich.xchangestream.kraken.dto.enums.KrakenOrderBookMessageType;
-import info.bitrich.xchangestream.service.netty.StreamingObjectMapperHelper;
 import org.apache.commons.lang3.StringUtils;
+import org.knowm.xchange.dto.marketdata.OrderBook;
 import org.knowm.xchange.kraken.dto.marketdata.KrakenPublicOrder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.math.BigDecimal;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Kraken order book utils
@@ -97,4 +92,26 @@ public class KrakenOrderBookUtils {
         return new BigDecimal(timestamp).multiply(new BigDecimal(1000)).longValue();
     }
 
+    public static OrderBook verifyKrakenOrderBook(OrderBook orderBook) {
+
+        Collections.sort(orderBook.getBids());
+        Collections.sort(orderBook.getAsks());
+
+        orderBook
+                .getAsks()
+                .removeIf(
+                        ask ->
+                                (orderBook.getBids().get(0).getLimitPrice().compareTo(ask.getLimitPrice()) >= 0
+                                        && ask.getTimestamp().before(orderBook.getBids().get(0).getTimestamp()))
+                );
+
+        orderBook
+                .getBids()
+                .removeIf(
+                        bid ->
+                                (orderBook.getAsks().get(0).getLimitPrice().compareTo(bid.getLimitPrice()) <= 0
+                                        && bid.getTimestamp().before(orderBook.getAsks().get(0).getTimestamp()))
+                );
+        return new OrderBook(orderBook.getTimeStamp(),orderBook.getAsks(),orderBook.getBids());
+    }
 }
--- a/xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingMarketDataService.java
+++ b/xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingMarketDataService.java
@@ -12,6 +12,7 @@ import org.knowm.xchange.dto.marketdata.OrderBook;
 import org.knowm.xchange.dto.marketdata.Ticker;
 import org.knowm.xchange.dto.marketdata.Trade;
 import org.knowm.xchange.kraken.KrakenAdapters;
+import org.knowm.xchange.kraken.KrakenUtils;
 import org.knowm.xchange.kraken.dto.marketdata.KrakenPublicOrder;
 import org.knowm.xchange.kraken.dto.marketdata.KrakenPublicTrade;
 import org.knowm.xchange.kraken.dto.marketdata.KrakenTicker;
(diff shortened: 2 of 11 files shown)


xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenOrderBookUtils.java
```diff
@@ -97,4 +92,26 @@ private static long timestampToMs(String timestamp) {
         return new BigDecimal(timestamp).multiply(new BigDecimal(1000)).longValue();
     }
 
+    public static OrderBook verifyKrakenOrderBook(OrderBook orderBook) {
+
+        Collections.sort(orderBook.getBids());
```
badgerwithagun: The sort looks wrong. Order books are usually sorted in descending order on one side and ascending on the other.

badgerwithagun:
What bug is this resolving?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingMarketDataService.java
```diff
@@ -57,7 +58,7 @@ public KrakenStreamingMarketDataService(KrakenStreamingService service) {
                 .map(ob -> {
                     KrakenOrderBookStorage orderBook = ob.toKrakenOrderBook(orderBooks.get(channelName), depth);
                     orderBooks.put(channelName, orderBook);
-                    return KrakenAdapters.adaptOrderBook(orderBook.toKrakenDepth(), currencyPair);
+                    return KrakenOrderBookUtils.verifyKrakenOrderBook(KrakenAdapters.adaptOrderBook(orderBook.toKrakenDepth(), currencyPair));
```
badgerwithagun: This all looks like it should occur inside `adaptOrderBook` BEFORE creation of the `OrderBook` object.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenOrderBookUtils.java
```diff
@@ -97,4 +92,26 @@ private static long timestampToMs(String timestamp) {
         return new BigDecimal(timestamp).multiply(new BigDecimal(1000)).longValue();
     }
 
+    public static OrderBook verifyKrakenOrderBook(OrderBook orderBook) {
```
badgerwithagun: General points on design:

This method has a functional-style signature but has side effects.

`Collections.sort` and `removeIf` both mutate the existing lists, so you are doing multiple things:

1. Sorting the _existing_ order book
1. Modifying the_ existing_ order book to remove entries
1. Creating a new order book with the modified entry lists (leaving two identical `OrderBook` objects containing the same entries).

Instead, it would make more sense to either create the new `Orderbook` using a filtered copy and sorted of the original (using a consistent sort) OR return `void` and explicitly modify the existing order book.

The method is called `verify...` so I'd usually expect it to check something.  Would `sortAndRemoveInvalidEntries` make more sense?

In any case, I think this code is in the wrong place. It should happen in `adaptOrderBook` _before_ creating the `OrderBook`.

I'm still a little unclear on what issue this is resolving, too.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingMarketDataService.java
makarid: I would have put it on adaptOrderbook but we have some version issues and that is why i thought it would be better to have it on bitrich-info project instead.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenOrderBookUtils.java
makarid: OK, i would just return a void. The issue that this method fix is that a lot of times the stream doesn't send some messages and there are times that the orderbook is not consistent with the real Kraken orderbook. I use this implementation on my own project and it works very well. I have found the same issue on CoinbaseProOrderbook also and i will try to fix this too.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenOrderBookUtils.java
makarid: Will this : Collections.sort(orderBook.getBids(),Collections.reverseOrder()); fix the issue with sorting?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenOrderBookUtils.java
makarid: I believe that there is a general issue with inconsistent orderbooks in this library, i have encounter issues on Bitmex,Kraken,CoinbasePro exchange. I don't use any REST API calls to get a new snapshot. As i have seen in issue #387 , there are more exchanges with this inconsistent orderbook thing. Maybe we need to discuss about implement a global approach in this, maybe we should combine a websocket stream with REST API, or at least create a validation method in order to delete all the inconsistent orderbook entries, like the verifyOrderbook method.

makarid:
The major issue on all exchanges which have the inconsistency orderbook is that some bids/asks prices does not exists but the orderbook still have them. Maybe the verifyOrderbook method is a little hacky and not a solution that really solves the issue. If it is true ,then we need to discuss about that and find what will be a better implementation. Until then, we can use the verifyOrderbook method in order to have a much more consistent orderbook, at least on the first 3-5 bid/ask prices.

makarid:
Well i just found the CoinbasePro issue which didn't delete the zero volume values. On the CoinbaseProWebsocketTransaction.CoinbaseProOrderbookChanges, in the line 138 we must change the way we check the volume !=0. Right now we check if !(String)volume.equals("0") which is not correct. We must change it to Double.parseDouble(volume) != 0.0 . I will implement it

pchertalev:
Hi @makarid 

I see you are trying to fix common problem - crossing asks and bids. :) 
We fixed this issue on application side: before using updated bitrich order book we always checks crossing book items and drop all of them (not only oldest item like you do in this pool request)
But I have the following concerns here. 
1. Fix of this issue will not allow you to detect corrupted orderbook on the application side. It looks like good idea, from my perspective, to request Kraken support about losing orderbook updates and fix issue on thr Kraken side.
2. Like u said It is VERY common problem for all exchanges and from my perspective it must be implemented on XChange side (not bitrich), it allowed to fix issue for all exchanges:
E.g. into constructor XChange OrderBook class: org.knowm.xchange.dto.marketdata.OrderBook

So I would suggest you to push this fix into XChange directly: https://github.com/knowm/XChange
Tim Molter is very active in XChange project and all pool requests are processed very quickly by him.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenOrderBookUtils.java
pchertalev: @makarid 
Always order book items are sorted by price: asks with ascending order and bids with descending order. See TreeSets here:  info/bitrich/xchangestream/kraken/KrakenOrderBookStorage.java:40
So you don't need to sort items before processing. 

If you have sorted bids in ascending order then order book became invalid in bids. 
So there is the question here did u test you changes before pool request created :)

PR #399, opened 2019-09-08
Get entire Cex.IO order book, handle sequence number mismatch smoothly

BenjaminTrapani:
Fixes an issue where the market data reader would crash due to trying to resubscribe to too many channels. Also makes entire order book available by default.


xchange-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingMarketDataService.java
```diff
@@ -39,17 +39,15 @@ public OrderBookUpdateConsumer(CexioStreamingRawService streamingOrderDataServic
...
-      if (prevID == null || prevID.add(BigInteger.ONE).compareTo(t.id) == 0) {
-        prevID = t.id;
-        orderBookSoFar = CexioAdapters.adaptOrderBookIncremental(orderBookSoFar, t);
-        retVal = orderBookSoFar;
-      } else {
-        streamingOrderDataService.resubscribeChannels();
-        retVal = orderBookSoFar;
-        orderBookSoFar =
+      if (prevID != null && prevID.add(BigInteger.ONE).compareTo(t.id) != 0) {
+        OrderBook orderBookSoFar =
```
badgerwithagun: This line doesn't look like it does anything! Just assigns a local variable which is never used.

BenjaminTrapani: Lol good catch, that's what I get for copy-pasting the assignment in the declaration. The statement is supposed to assign to the member variable, fixed.

badgerwithagun:
@dozd, the webhook from Travis doesn't seem to have fired and GitHub won't let me merge this. Would you mind?

badgerwithagun:

This needs @dozd's attention.

PR #400, opened 2019-09-09
Depend on 4.4.0 xchange

BenjaminTrapani:
Doing so makes it easier to fix bugs in xchange that manifest in xchange-stream.


badgerwithagun:
Generally in the past we've only advanced to release versions (so that'd be 21 right now).

Is there anything in 22-SNAPSHOT you particularly need?

BenjaminTrapani:
See https://github.com/knowm/XChange/pull/3208. That bug renders cexio basically unusable. Developing and testing the fix required using the snapshot version, so I figured I'd submit the changes here. Plus, developing against the xchange snapshot releases will help us more easily align our releases with xchange's.

makarid:
Totally agree with @BenjaminTrapani if it can be done. It will save a lot of headaches.

badgerwithagun:
I've bumped this PR to XChange 4.4.0 and will merge once green.

Issue #356, opened 2019-07-15
Kraken WebSocket reconnection leading to out of memory exception


jurepetrovic:
Hello, 

yesterday (14 July 2019) Kraken had official outage. 
The websocket service was disconnected and xchange-stream continously 
wanted to reconnect in short time intervals. 

A huge number of reconnection attempts lead to the following exception: 

2019-07-14 21:08:46 | ERROR | [nioEventLoopGroup-675-1] i.b.x.s.n.WebSocketClientHandler:66 - WebSocket Client failed to connect. Invalid handshake response getStatus: 503 Service Temporarily Unavailable
2019-07-14 21:08:47 | WARN  | [nioEventLoopGroup-675-1] i.b.x.k.KrakenStreamingService:211 - Problem with connection: class io.netty.handler.codec.http.websocketx.WebSocketHandshakeException - Invalid handshake response getStatus: 503 Service Temporarily Unavailable
2019-07-14 21:08:47 | WARN  | [nioEventLoopGroup-681-1] i.b.x.k.KrakenStreamingService:213 - Problem with connection
java.lang.IllegalStateException: executor not accepting a task
	at io.netty.resolver.AddressResolverGroup.getResolver(AddressResolverGroup.java:60)
	at io.netty.bootstrap.Bootstrap.doResolveAndConnect0(Bootstrap.java:200)
	at io.netty.bootstrap.Bootstrap.access$000(Bootstrap.java:49)
	at io.netty.bootstrap.Bootstrap$1.operationComplete(Bootstrap.java:188)
	at io.netty.bootstrap.Bootstrap$1.operationComplete(Bootstrap.java:174)
	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:502)
	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:476)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:415)
	at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:540)
	at io.netty.util.concurrent.DefaultPromise.setSuccess0(DefaultPromise.java:529)
	at io.netty.util.concurrent.DefaultPromise.trySuccess(DefaultPromise.java:101)
	at io.netty.channel.DefaultChannelPromise.trySuccess(DefaultChannelPromise.java:84)
[...]

PR #410, opened 2019-09-18
#356 Fix of the netty memory leak in case target exchange is broken

pchertalev:
There is double resubscribing in case exchange socket has been broken

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -202,7 +202,6 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
                                         completable.onComplete();
                                     } else {
                                         webSocketChannel.disconnect().addListener(x -> {
-                                            scheduleReconnect();
                                             completable.onError(handshakeFuture.cause());
                                         });
                                     }


service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -202,7 +202,6 @@ protected void initChannel(SocketChannel ch) {
                                         completable.onComplete();
                                     } else {
                                         webSocketChannel.disconnect().addListener(x -> {
-                                            scheduleReconnect();
```
badgerwithagun: Nice! Just so I can follow the logic, which is the second reconnect that we hit?

pchertalev: https://github.com/bitrich-info/xchange-stream/blob/6c2486cd3429e2eb15f76fec334318bd5c119e24/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java#L435

badgerwithagun:
Ok, this looks good. Fine work for digging this one out @pchertalev.

pchertalev:
Not at all. This issue was spoiling my life to long. :) Thank you for quick merging!

PR #412, opened 2019-09-24
Poloniex orderbook consistency

davidjirovec:
I've noticed sometimes also Poloniex implementation has problem with bid price > ask price. Cause of this issue here (but maybe also in other exchange implementations) is usage of flatMap on incoming orderbook update which contains multiple changes. The message should be treated as one atomic update, instead multiple org.knowm.xchange.dto.marketdata.OrderBook instances were emmited, some inconsistent. This seems to be fixed now, only one and consistent instance of org.knowm.xchange.dto.marketdata.OrderBook is emmited. I guess performance of applications using this library could be improved because less messages need to be processed. Other changes include

- Use rxjava scan operator instead of shared map of orderbooks. To make the implementation completely safe, info.bitrich.xchangestream.poloniex2.dto.PoloniexOrderbook should be immutable.
- Check sequence numbers of incoming messages. Api documentation mentions `number increasing by one with each update and is unique within its channel`, so I've added check for this. Question would be what to do when sequences are broken. I just throw an exception, but this never happened anyway.


xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
```diff
@@ -75,10 +81,27 @@ public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<
     @Override
     public Observable<Trade> getTrades(CurrencyPair currencyPair, Object... args) {
         Observable<PoloniexWebSocketTradeEvent> subscribedTrades = service.subscribeCurrencyPairChannel(currencyPair)
+                .flatMapIterable(poloniexWebSocketEvents -> poloniexWebSocketEvents)
                 .filter(s -> s.getEventType().equals("t"))
```
badgerwithagun: Might be slightly clearer if this were:
```
.filter(PoloniexWebSocketTradeEvent.class::isInstance)
```
Since we've gone to the effort of assigning these types beforehand.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
```diff
@@ -39,22 +46,21 @@ public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<
...
-                        orderbook = orderbooks.get(currencyPair);
-                        orderbook.modify(modifiedEvent);
-                    }
-                    return orderbook;
-                });
+                .scan(
+                        Optional.empty(),
+                        (Optional<PoloniexOrderbook> orderbook, List<PoloniexWebSocketEvent> poloniexWebSocketEvents) ->
+                                poloniexWebSocketEvents.stream()
+                                        .filter(s -> s.getEventType().equals("i") || s.getEventType().equals("o"))
```
badgerwithagun:
```
.filter(s - > PoloniexWebSocketOrderbookInsertEvent.class.isInstance(s) || PoloniexWebSocketOrderbookModifiedEvent.class.isInstance(s))
```
would be clearer I think

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
```diff
@@ -39,22 +46,21 @@ public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<
     @Override
     public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
         Observable<PoloniexOrderbook> subscribedOrderbook = service.subscribeCurrencyPairChannel(currencyPair)
```
badgerwithagun: All three of these methods are creating duplicate fresh subscriptions to the same channel, so they are all doing the same work in `PoloniexStreamingService` and `PoloniexWebSocketEventsTransaction`.  It'd be nice if we could re-use those `service.subscribeCurrencyPairChannel` subscriptions, since all that Jackson work is quite expensive.  Could we cache them by `CurrencyPair`?

It's unrelated, so feel free to just say "maybe another time" and I'll shut up.  :)

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketEventsTransaction.java
```diff
@@ -19,14 +20,37 @@
...
+
+    public Long getSeqId() {
+        return seqId;
+    }
+
+    public List<PoloniexWebSocketEvent> getEvents() {
+        return events;
+    }
+
+    private List<PoloniexWebSocketEvent> createEvents(final List<JsonNode> jsonEvents) {
```
badgerwithagun: Maybe not one for now, but I'm pretty sure all this work can be done by adding a Jackson deserializer annotation to `PoloniexWebSocketEvent` which causes the different event types to be deserialized as the concrete classes.

See the changes on https://github.com/bitrich-info/xchange-stream/pull/385/commits/7059acb767945daf35fce743ddc097e0b350438e to `LgoOrderEvent`

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
davidjirovec: True, but I've used `instanceof`, IntelliJ did not like this usage of `class.isInstance`

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
davidjirovec: Yes, maybe another time ;) At least the JsonNode subscriptions are still reused using `info.bitrich.xchangestream.poloniex2.PoloniexStreamingService#subscriptions`

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketEventsTransaction.java
davidjirovec: I've originally tried to make the deserialization based purely on Jackson annotations, but had some problems, don't remember now, sorry.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
badgerwithagun: One day I will try and understand IntelliJ. For now I consider it voodoo and stick to Eclipse like a caveman.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketEventsTransaction.java
badgerwithagun: No problem. This is fine.

badgerwithagun:
Tests had failed on Bitmex, which I think is intermittent, so I've kicked off the build again.

PR #422, opened 2019-10-08
[coinbase-pro] Fix corruption of orderbooks & improve performance

badgerwithagun:
Noticed that I was getting zeros appearing in CBP orderbooks. This turned out to be due to concurrency issues with the use of `HashMap` to store the state.

Also made the `getOrderBook` method support any `Number` type for the max depth.

Finally fixed very high CPU usage caused by multiple format translations.

### DIFF
--- a/xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingAdapters.java
+++ b/xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingAdapters.java
@@ -1,5 +1,11 @@
 package info.bitrich.xchangestream.coinbasepro;
 
+import java.math.BigDecimal;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.TimeZone;
+
 import org.knowm.xchange.coinbasepro.CoinbaseProAdapters;
 import org.knowm.xchange.coinbasepro.dto.trade.CoinbaseProOrder;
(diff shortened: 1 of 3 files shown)


mdvx:
Test of this patch passes - will close #424 and #426

jkolobok:
Can you fix "out-of-date" problem?

badgerwithagun:
@mdvx - that code was copied out of XChange. I'll tidy it up.

badgerwithagun:
@jkolobok - this fixes all the issues I've found with the CBP orderbook.

badgerwithagun:
@mdvx - I'm not aware of a way of telling if a `BigDecimal` is equivalent to zero without running it through a number conversion to do so. Since any zeros will be filtered out at line 146 they will only get created in the eden space and immediately cleared out so it's probably efficient enough.

mdvx:
> @mdvx - I'm not aware of a way of telling if a `BigDecimal` is equivalent to zero without running it through a number conversion to do so. Since any zeros will be filtered out at line 146 they will only get created in the eden space and immediately cleared out so it's probably efficient enough.

some of the APIs call (different number of 0 for different exchanges) so save on evaluating the string
"0.000000".equals(newVal) ? return BigDecimal.ZERO : new BigDecimal(newVal);
it only works when the exchange always returns "0.000000" as a fixed string always meaning zero.  But your right, it is just eden space

badgerwithagun:
@dozd

badgerwithagun:
Unfortunately until either @dozd or @Flemingjp review this, I can't merge it, since I wrote it.

PR #424, opened 2019-10-18
NPE at line 130

mdvx:
after disconnecting and re-connectingto the exchange, I am getting L2UPDATES before SNAPSHOT updates for a pair.


badgerwithagun:
I have a feeling this is fixed by #422 @mdvx - could you try it?

mdvx:
I will apply #422 it, and close this #424 and #426 in preference to #422 (Direct to XChange format was my next goto)

mdvx:
use #426 instead

PR #437, opened 2019-10-28
POM change: Java 9+ support (JAXB is missing now)

arass:
JAXB has been removed from Java9 and higher. Need to add it as a library going forward. Quick change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,6 +117,13 @@
             <artifactId>rxjava</artifactId>
             <version>2.2.13</version>
         </dependency>
+        
+   		<dependency>
+			<groupId>javax.xml.bind</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -117,6 +117,13 @@
             <artifactId>rxjava</artifactId>
             <version>2.2.13</version>
         </dependency>
+        
+   		<dependency>
```
badgerwithagun: Hi @arass! Thanks for the contribution. Seems reasonable.

Can you just line up the indentation with the rest of the file and bump the version up to the latest (2.3.2)?

badgerwithagun:
Odd. My code's pulling in 2.3.2. Oh well, Dependabot will fix it if there is a later one.

Thanks!

Issue #133, opened 2018-03-15
Trade Type is always ASK on Poloniex2


stepanurban:
Hi all,
I use simple example for LTC/BTC on Ploniex
exchange.getStreamingMarketDataService().getTrades(pair).subscribe(trade -> {
            System.out.println(trade);
        }, throwable -> LOG.error("ERROR in getting trades: ", throwable));

and I receive trade Type always ASK
Trade [type=ASK, originalAmount=0.06207470, currencyPair=LTC/BTC, price=0.02003588, timestamp=Thu Mar 15 13:48:39 CET 2018, id=15433354]
Trade [type=ASK, originalAmount=2E-8, currencyPair=LTC/BTC, price=0.02004129, timestamp=Thu Mar 15 13:48:39 CET 2018, id=15433355]

On Poloniex website there is true value
2018-03-15 12:48:39	Buy	0.02004129	0.00000002	0.00000000
2018-03-15 12:48:39	Sell	0.02003588	0.06207470	0.00124372
Can it be corrected in some way

Thanks
Stepan

Flemingjp:
@stepanurban Can you check if this still occurs with the 4.3.11 release?

jkolobok:
still occurs in 4.3.16

PR #438, opened 2019-11-02
#133 Fixed poloniex trade parsing from websocket

jkolobok:
It was using rest api entities while websockets use different type

### DIFF
--- a/xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
+++ b/xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
@@ -2,14 +2,7 @@ package info.bitrich.xchangestream.poloniex2;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import info.bitrich.xchangestream.core.StreamingMarketDataService;
-import info.bitrich.xchangestream.poloniex2.dto.OrderbookInsertEvent;
-import info.bitrich.xchangestream.poloniex2.dto.OrderbookModifiedEvent;
-import info.bitrich.xchangestream.poloniex2.dto.PoloniexOrderbook;
-import info.bitrich.xchangestream.poloniex2.dto.PoloniexWebSocketEvent;
-import info.bitrich.xchangestream.poloniex2.dto.PoloniexWebSocketOrderbookInsertEvent;
-import info.bitrich.xchangestream.poloniex2.dto.PoloniexWebSocketOrderbookModifiedEvent;
-import info.bitrich.xchangestream.poloniex2.dto.PoloniexWebSocketTickerTransaction;
-import info.bitrich.xchangestream.poloniex2.dto.PoloniexWebSocketTradeEvent;
+import info.bitrich.xchangestream.poloniex2.dto.*;
 import info.bitrich.xchangestream.service.netty.StreamingObjectMapperHelper;
 import io.reactivex.Observable;
 import org.knowm.xchange.currency.CurrencyPair;
@@ -20,13 +13,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.math.BigDecimal;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.SortedMap;
+import java.util.*;
 
 import static org.knowm.xchange.poloniex.PoloniexAdapters.adaptPoloniexDepth;
-import static org.knowm.xchange.poloniex.PoloniexAdapters.adaptPoloniexPublicTrade;
 import static org.knowm.xchange.poloniex.PoloniexAdapters.adaptPoloniexTicker;
 
 /**
@@ -90,12 +79,26 @@ public class PoloniexStreamingMarketDataService implements StreamingMarketDataSe
                 .share();
 
         return subscribedTrades
-                .map(s -> adaptPoloniexPublicTrade(s.toPoloniexPublicTrade(currencyPair), currencyPair));
+                .map(s -> adaptPoloniexPublicTrade(s, currencyPair));
     }
 
-    private Optional<PoloniexOrderbook> getPoloniexOrderbook(
-            final Optional<PoloniexOrderbook> orderbook, final PoloniexWebSocketEvent s
-    ) {
+    private static Trade adaptPoloniexPublicTrade(
+            PoloniexWebSocketTradeEvent poloniexTradeEvent, CurrencyPair currencyPair) {
+        TradeEvent tradeEvent = poloniexTradeEvent.getTradeEvent();
+        Date timestamp = new Date(tradeEvent.getTimestampSeconds() * 1000);
+        Trade trade =
+                new Trade(
+                        tradeEvent.getType(),
+                        tradeEvent.getSize(),
+                        currencyPair,
+                        tradeEvent.getPrice(),
+                        timestamp,
+                        tradeEvent.getTradeId());
+        return trade;
+    }
+
+    private Optional<PoloniexOrderbook> getPoloniexOrderbook(final Optional<PoloniexOrderbook> orderbook,
+                                                             final PoloniexWebSocketEvent s) {
         if (s.getEventType().equals("i")) {
             OrderbookInsertEvent insertEvent = ((PoloniexWebSocketOrderbookInsertEvent) s).getInsert();
             SortedMap<BigDecimal, BigDecimal> asks = insertEvent.toDepthLevels(OrderbookInsertEvent.ASK_SIDE);
--- a/xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
+++ b/xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
@@ -79,22 +79,7 @@ public class PoloniexStreamingMarketDataService implements StreamingMarketDataSe
                 .share();
 
         return subscribedTrades
-                .map(s -> adaptPoloniexPublicTrade(s, currencyPair));
-    }
-
-    private static Trade adaptPoloniexPublicTrade(
-            PoloniexWebSocketTradeEvent poloniexTradeEvent, CurrencyPair currencyPair) {
-        TradeEvent tradeEvent = poloniexTradeEvent.getTradeEvent();
-        Date timestamp = new Date(tradeEvent.getTimestampSeconds() * 1000);
-        Trade trade =
-                new Trade(
-                        tradeEvent.getType(),
-                        tradeEvent.getSize(),
-                        currencyPair,
-                        tradeEvent.getPrice(),
-                        timestamp,
-                        tradeEvent.getTradeId());
-        return trade;
+                .map(s -> PoloniexWebSocketAdapter.convertPoloniexWebSocketTradeEventToTrade(s, currencyPair));
     }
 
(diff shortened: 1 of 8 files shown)


xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
```diff
@@ -90,12 +79,26 @@ public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<
...
 
         return subscribedTrades
-                .map(s -> adaptPoloniexPublicTrade(s.toPoloniexPublicTrade(currencyPair), currencyPair));
+                .map(s -> adaptPoloniexPublicTrade(s, currencyPair));
     }
 
-    private Optional<PoloniexOrderbook> getPoloniexOrderbook(
-            final Optional<PoloniexOrderbook> orderbook, final PoloniexWebSocketEvent s
-    ) {
+    private static Trade adaptPoloniexPublicTrade(
```
badgerwithagun: Could you move this to a static method on a `PoloniexAdapters`class?

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
```diff
@@ -90,12 +79,26 @@ public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<
...
+            PoloniexWebSocketTradeEvent poloniexTradeEvent, CurrencyPair currencyPair) {
+        TradeEvent tradeEvent = poloniexTradeEvent.getTradeEvent();
+        Date timestamp = new Date(tradeEvent.getTimestampSeconds() * 1000);
+        Trade trade =
+                new Trade(
```
badgerwithagun: Better to use `Trade.Builder`

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketEventsTransaction.java
```diff
@@ -53,31 +50,50 @@ public Long getSeqId() {
...
+                    int timestampSeconds = jsonNode.get(5).asInt();
+                    TradeEvent tradeEvent = new TradeEvent(tradeId,
+                            side,
+                            price,
+                            size,
+                            timestampSeconds);
+                    PoloniexWebSocketTradeEvent insertEvent = new PoloniexWebSocketTradeEvent(tradeEvent);
+                    events.add(insertEvent);
+                    break;
+                }
```
badgerwithagun: Needs a `default` case

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/TradeEvent.java
```diff
@@ -9,48 +9,36 @@
...
 public class TradeEvent {
     private String tradeId;
-    private String type;
-    private BigDecimal rate;
-    private BigDecimal amount;
-    private String date;
+    private Order.OrderType type;
+    private BigDecimal price;
+    private BigDecimal size;
+    private int timestampSeconds;
```
badgerwithagun: You're sort of mixing source-style and target-style data here.  For clarity, it'd be better if:

* EITHER all these properties were in the source format (as they were before) and all converted when the object is mapped to an XChange object later
* OR all the properties were in the target format (in which case this field should be a `Date` rather than an `int`).

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
jkolobok: No, because it's a different library. We can create a different class, but moving it to xchange doesn't make sense.

jkolobok: though I'll rename the method. as it doesn't reflect what's going on

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
jkolobok: why? I think builder is redundant here because all fields are always required... so you don't have any flexibility in building, thus don't need Builder

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketEventsTransaction.java
jkolobok: And what to do there? default is noop. we cann't throw an exception there.. if they add new type we should just ignore it

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/TradeEvent.java
jkolobok: I agree, but partially. Usage of side as "1" or "0" is very difficult, so it makes sense to convert it to something clearly understandable  asap to avoid confusion later in the code (which actually caused this bug). For example in REST api it comes as string "buy" or "sell" which caused incorrect handling.  So I treat this value still as SOURCE (in your terms).

Regarding the Date vs int. Date is different for example it has precision of milliseconds. I actually think that Date is  bad choice for timestamp, but it's in the core library so we are not going to mess with it. 
Originally the code was converting int->String then String -> Date. So here imo we also use Source value and later convert .
We can move conversion to TradeEvent and let it return both original value(string, int) and  "adapted" (OrderType, Date).
Though I think that storing "1" and "0" is a problem and will cause confusion, so we should drop them asap

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
badgerwithagun: Because then the code doesn't get broken when new fields are added to `Trade`, as has happened several times.

Also,IIRC walec intends to remove the publicly visible constructors in XChange in the future.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
badgerwithagun: I didn't mean moving it to XChange. I just meant moving it to a utility class, making it default-visible.  See `BitfinexStreamingAdapters` for an example.

Once there, you can/could write a unit test for it (which I'm not asking you to write, because I'm trying to avoid creating too much extra work for you, but at least it means someone else could in the future and it keeps this class tidy).

jkolobok: ok. I'll create a new class for the conversion code

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
jkolobok: maybe is should get broken for example if new fields are required?:) I suggest we create a ticket to make an improvement in all places instead of just this one and implement it later. What do you think?

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketEventsTransaction.java
badgerwithagun: On my IDE, and on Sonar analysis, this throws a warning due to the lack of a default condition. It's a code smell. You should be explicit that you are _deliberately_ ignoring the no-op condition:

```
  default: // Ignore
```

This is an [age-old debate](https://stackoverflow.com/questions/4649423/should-switch-statements-always-contain-a-default-clause?lq=1) and I'd rather not open it again. The [Google Style Guide](https://google.github.io/styleguide/javaguide.html#s4.8.4-switch) insists on it, as do most projects I work on, and it's not really much to ask.

jkolobok: ok

jkolobok: is there a page about contributions? let's add a note about google style guide there. Cause for me it's a question of IDE preferences.  I don't like redundant code, but if it's a local rule that's fine.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
badgerwithagun: I'm happy with just progressive improvement, but if you felt like doing a widespread cleanup PR that would obviously be really appreciated!

badgerwithagun:

This needs @dozd's attention.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketEventsTransaction.java
badgerwithagun: There isn't a style guide for this repo, but I have been thinking we should add the coveo fmt plugin like XChange and start enforcing Google guidelines, since there is a plan to merge with XChange anyway and it would be good preparatory work.

See long discussion here: https://github.com/knowm/XChange/issues/3220

jkolobok: Then we definitely need a separate ticket for this. It's going to be a lot of work anyway. let's keep PRs more atomic:)

jkolobok: created a ticket for this https://github.com/bitrich-info/xchange-stream/issues/441

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
jkolobok: fixed

jkolobok:
> Please note that I am unable to merge PRs at present due to an issue with Travis not notifying GitHub that the build ran successfully. The repository is set to disallow merges if checks are in progress/failing so merging is disabled.
> 
> This needs @dozd's attention.

Apparently checks have passed

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketEventsTransaction.java
badgerwithagun: Just needs this default case

jkolobok:
Failed on bitmex after merging dev. Can we just rerun it?

PR #440, opened 2019-11-02
 check if bids/asks contains ccy

mdvx:
in-case we did not get l2update before snapshot - this happens during reconnects

### DIFF
--- a/xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingMarketDataService.java
+++ b/xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingMarketDataService.java
@@ -63,7 +63,8 @@ public class CoinbaseProStreamingMarketDataService implements StreamingMarketDat
         return getRawWebSocketTransactions(currencyPair, false)
                 .filter(message -> message.getType().equals(SNAPSHOT) || message.getType().equals(L2UPDATE))
                 .map(s -> {
-                    if (s.getType().equals(SNAPSHOT)) {
+                    // check if bids/asks contains ccy - in-case we did not get the snapshot (this happens during reconnects)
+                    if (s.getType().equals(SNAPSHOT) || !bids.containsKey(currencyPair) || !asks.containsKey(currencyPair)) {
                         bids.put(currencyPair, new TreeMap<>(java.util.Collections.reverseOrder()));
                         asks.put(currencyPair, new TreeMap<>());
                     }


mdvx:
further investigation has shown that this occured when connect was called before disconnect completed, in a disconnect/(re)connect cycle

badgerwithagun:

This needs @dozd's attention.

dozd:
I do not know why it stopped working... Trying to restart the builds there.

dozd:
@badgerwithagun seems to be fixed now

xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingMarketDataService.java
```diff
@@ -63,7 +63,8 @@ private boolean containsPair(List<CurrencyPair> pairs, CurrencyPair pair) {
         return getRawWebSocketTransactions(currencyPair, false)
                 .filter(message -> message.getType().equals(SNAPSHOT) || message.getType().equals(L2UPDATE))
                 .map(s -> {
-                    if (s.getType().equals(SNAPSHOT)) {
+                    // check if bids/asks contains ccy - in-case we did not get the snapshot (this happens during reconnects)
+                    if (s.getType().equals(SNAPSHOT) || !bids.containsKey(currencyPair) || !asks.containsKey(currencyPair)) {
```
badgerwithagun: Isn't this just going to result in us writing a partial orderbook for the pair?  I wonder if it might be better just to _ignore_ (filter out) updates like these until we get a valid snapshot.

mdvx: yes, i will produce a partial book until SNAPSHOT arrives.  I am good with ignoring too, my goal was to avoid the NPE that occured later down the line, when only one side of the book was present.   Also, I need to take another run thru this code as I am getting some strange results with crossed books, where the L2UPDATE is the only content of the book.  it may be the SNAPSHOT never arrives.

badgerwithagun: OK.  Yes - I think I prefer ignoring and not starting to send orderbooks to the client until we have a clean snapshot, so if you could change that, great.

Good luck finding the other issue.

mdvx: As I stand today, we dont get a clean SNAPSHOT, it gets losts and L2UPDATES pour in.  Is it possible the caching mech is allowing that?, it only seems to hold one last message.
@badgerwithagun

mdvx:
No longer needed, something else fixed it, or i started using the api better

xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingMarketDataService.java
mdvx: closing

PR #447, opened 2019-11-06
applyStreamingSpecification to CoinbasePro

mdvx:
So we can pick up and use SOCKS_PROXY_HOST/SOCKS_PROXY_PORT etc...


mdvx:
No, but it is a start, and it does also applies the httpReadTimeOut and setHttpConnTimeout- i think all exchanges need this. go some distance to getting Netty working with the Proxy, but not there yet

PR #3, opened 2017-04-11
GDAX Exchange support proposal

lucarosellini:
This PR add support for GDAX WebSocket feed. I know there is room for improvement. Any feedback is welcome.

PR #457, opened 2019-11-28
[Kraken] private channel

mdvx:
A first functional pass at Kraken private channel
Known Issues
* getOrderChanges & getUserTrades ignore currencyPair parameter (and subscribe to all)
** now does its best to filter
* does not unsubscribe from channels (ownTrades & openOrders)

UnKnown Issues
* possibily some


badgerwithagun:
Is this ready for review, @mdvx ?

mdvx:
Yes, lets do a first review - please

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -18,47 +17,72 @@
...
+    private KrakenStreamingTradeService streamingTradeService;
+
     private final SynchronizedValueFactory<Long> nonceFactory = new CurrentTimeNonceFactory();
 
     public KrakenStreamingExchange() {
-        this.streamingService = new KrakenStreamingService(API_URI);
+    }
+
+    private static String pickUri(boolean isPrivate, Boolean useSandbox) {
+        if (useSandbox != null && useSandbox)
```
badgerwithagun: Is the beta environment really a "sandbox"?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -18,47 +17,72 @@
...
+            return API_BETA_URI;
+        else
+            return isPrivate ? API_AUTH_URI : API_URI;
     }
 
     @Override
     protected void initServices() {
         super.initServices();
-        streamingMarketDataService = new KrakenStreamingMarketDataService(streamingService);
+        Boolean useSanbox = (Boolean)exchangeSpecification.getExchangeSpecificParametersItem(USE_SANDBOX);
```
badgerwithagun: `useSandbox` (spelling).

Also, if you use `MoreObjects.firstNonNull((Boolean)exchangeSpecification.getExchangeSpecificParametersItem(USE_SANDBOX), false)` you can avoid downstream null checks

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -18,47 +17,72 @@
...
+
+        this.streamingService = new KrakenStreamingService(false, pickUri(false,useSanbox));
+        this.streamingMarketDataService = new KrakenStreamingMarketDataService(streamingService);
+
+        if (exchangeSpecification.getApiKey() != null) {
```
badgerwithagun: Use `StringUtils.isNotEmpty`; when substituting environment variables into application config, it's hard to force a configuration value to null.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -18,47 +17,72 @@
...
     @Override
     public boolean isAlive() {
-        return streamingService.isSocketOpen();
+        return streamingService.isSocketOpen() && (privateStreamingService == null || privateStreamingService.isSocketOpen());
     }
     
     @Override
     public Observable<Object> connectionSuccess() {
-        return streamingService.subscribeConnectionSuccess();
+        return streamingService.subscribeConnectionSuccess().mergeWith(privateStreamingService.subscribeConnectionSuccess());
```
badgerwithagun: Hmmm. It seems a little unexpected that you'll get two connection successes. I wonder what people use this observable for?  With this possibility, you can't rely on it for state on its own; you need to check `isAlive()` as well, which should probably documented if we think this is the right behaviour.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -18,47 +17,72 @@
...
     }
     
     @Override
     public Observable<Throwable> reconnectFailure() {
-        return streamingService.subscribeReconnectFailure();
+        return streamingService.subscribeReconnectFailure().mergeWith(privateStreamingService.subscribeReconnectFailure());
```
badgerwithagun: Same applies here, possibly

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -27,18 +27,20 @@
  * @author makarid, pchertalev
  */
 public class KrakenStreamingService extends JsonNettyStreamingService {
```
badgerwithagun: Looking at this code, there seems to be so many `if.. else` sections, I wonder if it might be more clearly implemented as two separate classes, one private and one public. Thoughts?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
```diff
@@ -0,0 +1,221 @@
...
+            token = rawKrakenAcctService.getKrakenWebsocketToken();
+            tokenExpires = System.currentTimeMillis() + (token.getExpiresInSeconds() * 900L); // 900L instead of a 1000L for 90%
+        }
+        return token;
+    }
+
+    @Override
+    public Observable<Order> getOrderChanges(CurrencyPair currencyPair, Object... args) {
+
+        return Observable.create( source -> {
```
badgerwithagun: Why the wrapper observable? Looks like you can just return the result of `subscribeChannel`. Something like this (probably doesn't work, but certainly something like this):
```
return streamingService.subscribeChannel(getChannelName(KrakenSubscriptionName.openOrders), renewToken().getToken())
    .filter(JsonNode::isArray)
    .filter(Objects::nonNull)
    .map(jsonNode -> jsonNode.get(0))
    .map(jsonNode -> StreamingObjectMapperHelper.getObjectMapper().treeToValue(jsonNode, KrakenDtoOrderHolder[].class))
    .flatMapIterable(this::adaptKrakenOrders);
```

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
```diff
@@ -0,0 +1,221 @@
...
+                                adaptKrakenOrders(list)
+                                        .forEach(source::onNext));
+
+            } catch (Throwable e) {
+                source.onError(e);
+            }
+        });
+    }
+
+    private List<Order> adaptKrakenOrders(KrakenDtoOrderHolder[] dtoList) {
```
badgerwithagun: If you changed this method to return `Iterable<Order>` you could do the conversion using an Guava `FluentIterable` and avoid creating a temporary list. Minor suggestion, not required to pass review.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
```diff
@@ -0,0 +1,221 @@
...
+                Order.OrderType side = descr == null ? null : KrakenAdapters.adaptOrderType(KrakenType.fromString(descr.type));
+
+                Date time = dto.opentm == null ? null : new Date((long) (dto.opentm * 1000L));
+                String ordertype = (descr == null || descr.ordertype == null) ? null : descr.ordertype;
+
+                Set<Order.IOrderFlags> flags = adaptFlags(dto.oflags);
+
+                if ("limit".equals(ordertype))
+                    result.add(
+                            new LimitOrder.Builder(side, pair)
```
badgerwithagun: Most of these blocks are duplicated.  `LimitOrder.Builder` extends `Order.Builder` so you can create the builder in this `if` block, filling in the order-type specific bits, and then fill in all the common parts afterwards.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
```diff
@@ -0,0 +1,221 @@
...
+
+        for (String flag : arr) {
+            flags.add(KrakenOrderFlags.fromString(flag));
+        }
+        return flags;
+    }
+
+    @Override
+    public Observable<UserTrade> getUserTrades(CurrencyPair currencyPair, Object... args) {
+        return Observable.create( source -> {
```
badgerwithagun: Ditto - seems unnecessary to create a new `Observable` here

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: Not clear: but I changed the flag to Use_Beta
https://support.kraken.com/hc/en-us/articles/360000919926-Does-Kraken-offer-a-Test-API-or-Sandbox-Mode-

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
mdvx: It looks the next version  (beta-ws)  will merge the two back together (private and public), so might be worth waiting before creating new class for private?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
mdvx: no objection, it does not break anything.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
mdvx: fixed

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
mdvx: that works, and I added the missing currencyPair filter

mdvx: [Kraken] review fix #3
This is not great, we get updates without CurrencyPair, so we have to let those thru,
just in case they are for our subscription  (would have to maintain order cache otherwise)

                    .filter(order -> currencyPair == null || order.getCurrencyPair() == null || order.getCurrencyPair().compareTo(currencyPair) == 0);

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: Not sure what to do about this.  Wait for both, then emit one.  kind of like debounce.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: for reconnectFailure, I think it is OK to get two errors

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
badgerwithagun: I suppose that's probably the most predictable option, yes!

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
badgerwithagun: Ooo - what you could do here is have `streamingService` and `privateStreamingService` return different exception types?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
badgerwithagun: Fair enough.

xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
```diff
@@ -19,6 +19,8 @@
 public class CoinbaseProStreamingExchange extends CoinbaseProExchange implements StreamingExchange {
     private static final String API_URI = "wss://ws-feed.pro.coinbase.com";
     private static final String SANDBOX_API_URI = "wss://ws-feed-public.sandbox.pro.coinbase.com";
+    private static final String PRIME_API_URI = "wss://ws-feed.prime.coinbase.com";
```
badgerwithagun: Has this got mixed up with #476 ?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -18,46 +20,77 @@
...
         super.initServices();
-        streamingMarketDataService = new KrakenStreamingMarketDataService(streamingService);
+        Boolean useBeta = MoreObjects.firstNonNull((Boolean)exchangeSpecification.getExchangeSpecificParametersItem(USE_BETA), Boolean.FALSE);
+
+        this.streamingService = new KrakenStreamingService(false, pickUri(false,useBeta));
+        this.streamingMarketDataService = new KrakenStreamingMarketDataService(streamingService);
+
+        if (StringUtils.isNotEmpty(exchangeSpecification.getApiKey())) {
+            this.privateStreamingService = new KrakenStreamingService(true, pickUri(true,useBeta));
+//            this.futuresStreamingService = new KrakenStreamingService(true, API_FUTURES_URI);
```
badgerwithagun: Commented out code. Presumably this also means there's a compiler warning on the `API_FUTURES_API` constant.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: I like that, let me muse

xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
mdvx: yup

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: fixed

xchange-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
mdvx: fixed in commit d80358f69ab5f37c1d6d5d912e8d2b0ebb23e0d6

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: 558e6f61ed4922f3fea4625241a71f6a8fefac0d

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: 558e6f61ed4922f3fea4625241a71f6a8fefac0d

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -18,37 +22,61 @@
...
```
badgerwithagun: Best to remove this commented out code. Code related to this is also going to be needed in other places (`connect`/`disconnect` etc) so it's less confusing to leave it out and start a different PR to add futures.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -111,36 +136,98 @@ protected String getChannelNameFromMessage(JsonNode message) throws IOException
...
-        String pair = channelData[1];
         KrakenSubscriptionName subscriptionName = KrakenSubscriptionName.valueOf(channelData[0]);
 
-        int reqID = Math.abs(UUID.randomUUID().hashCode());
-        subscriptionRequestMap.put(reqID, channelName);
-        KrakenSubscriptionMessage subscriptionMessage = new KrakenSubscriptionMessage(reqID, KrakenEventType.unsubscribe,
-                Collections.singletonList(pair), new KrakenSubscriptionConfig(subscriptionName));
-        return objectMapper.writeValueAsString(subscriptionMessage);
+        if (isPrivate) {
+            String token = null; // TODO: We probably need to pass the token to unsubscribe (but we dont have args[0])
```
badgerwithagun: TODO

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -80,17 +81,11 @@ public boolean isAlive() {
     
     @Override
     public Observable<Object> connectionSuccess() {
-        if (privateStreamingService != null)
```
badgerwithagun: Why was this removed?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -80,17 +81,11 @@ public boolean isAlive() {
...
-            return streamingService.subscribeConnectionSuccess().mergeWith(privateStreamingService.subscribeConnectionSuccess());
-
         return streamingService.subscribeConnectionSuccess();
     }
     
     @Override
     public Observable<Throwable> reconnectFailure() {
-        if (privateStreamingService != null)
```
badgerwithagun: Why was this removed?

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: done

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: because I cannot resolve this, so going for the simplest option of just monitoring the public channel.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
mdvx: because I cannot resolve this, so going for the simplest option of just monitoring the public channel.

xchange-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
mdvx: removed

PR #460, opened 2019-12-04
Add connection idle to public API

davidjirovec:
Timeout values:

Bitfinex documentation says heartbeat is sent each 5 seconds. When testing this, heartbeat was coming each 15 seconds.

Poloniex mentions heartbeat each second and I can confirm this.

I've multiplied the timeout values x2 to be sure there are no false alarms, so Bitfinex 30 seconds, Poloniex 2 seconds.

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -1,20 +1,5 @@
 package info.bitrich.xchangestream.service.netty;
 
-import java.io.IOException;
-import java.net.URI;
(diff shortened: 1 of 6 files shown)


service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -53,12 +38,27 @@
 import io.reactivex.Observable;
 import io.reactivex.ObservableEmitter;
 import io.reactivex.subjects.PublishSubject;
+import org.slf4j.Logger;
```
badgerwithagun: We should really add coveo-fmt. Don't want to do it when there are a lot of outstanding PRs though. Catch-22.

Issue #463, opened 2019-12-06
Lots of compiler warnings needing cleanup prior to XChange merge


badgerwithagun:
The project has a large number of compiler warnings that should be cleaned up.  We should get those resolved prior to merging with XChange for quality reasons.

makarid:
I will do that

makarid:
Do i need to do anything more in order to close this issue?

PR #471, opened 2019-12-11
Fixing warnings 

makarid:
#463 . I managed to fix most compiler warnings. I am having trouble to replace the CertHelper.trustAllCerts() with whatever needs to replace. I do not understand how it works. If someone does i would appreciate to tell me.Thanks

One more thing is that PoloniexV1 has a weird list implementation which has annotation Beta and this brings some compiler warnings. I am not sure if it actually works because there is PoloniexV2 in the folder.

### DIFF
--- a/xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
+++ b/xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
@@ -73,7 +73,7 @@ public class PoloniexStreamingMarketDataService implements StreamingMarketDataSe
                                     bidQueue.removeIf(x -> rate.equals(x.getLimitPrice()));
                                 }
 
-                            } else if ("orderBookModify".equals(type)) {
+                            } else {
                                 if ("ask".equals(bookType)) {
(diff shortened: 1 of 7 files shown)


xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
```diff
@@ -32,7 +32,7 @@ public PoloniexStreamingMarketDataService(WampStreamingService streamingService)
     private Map<CurrencyPair, MinMaxPriorityQueue<LimitOrder>> orderBookAsks = new HashMap<>();
     private static final int ORDER_BOOK_LEVELS = 100;
 
-    Comparator<LimitOrder> asendingPriceComparator = (LimitOrder o1, LimitOrder o2) -> o1.getLimitPrice().compareTo(o2.getLimitPrice());
+    Comparator<LimitOrder> asendingPriceComparator = Comparator.comparing(LimitOrder::getLimitPrice);
     Comparator<LimitOrder> descendingPriceComparator = (LimitOrder o1, LimitOrder o2) -> -o1.getLimitPrice().compareTo(o2.getLimitPrice());
```
badgerwithagun: Pretty sure this can just be `asendingPriceComparator.reversed()`.

PR #477, opened 2019-12-18
Log the channel when connecting / idle / failed

mdvx:
So we can tell which exchange failed, when running multiple

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/WebSocketClientHandler.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/WebSocketClientHandler.java
@@ -51,7 +51,7 @@ public class WebSocketClientHandler extends SimpleChannelInboundHandler<Object> 
 
     @Override
     public void channelInactive(ChannelHandlerContext ctx) {
-        LOG.info("WebSocket Client disconnected!");
+        LOG.info("WebSocket Client disconnected! {}", ctx.channel());
     }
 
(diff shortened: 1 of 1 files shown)


service-netty/src/main/java/info/bitrich/xchangestream/service/netty/WebSocketClientHandler.java
```diff
@@ -51,7 +51,7 @@ public void channelActive(ChannelHandlerContext ctx) {
 
     @Override
     public void channelInactive(ChannelHandlerContext ctx) {
-        LOG.info("WebSocket Client disconnected!");
+        LOG.info("WebSocket Client disconnected! {}", ctx.channel());
```
badgerwithagun: What does this actually log?  I wasn't aware of anywhere that we give a useful name to the channel.

mdvx: As part of the compact log message, the ip address and hostname is in the string, with a couple of other less obvious details.  it really helps when you are connected to multiple exchanges.  (not currently in a position to give an example)

mdvx: example:
WebSocket Client connected! [id: 0x61a71f24, L:/192.168.0.100:57663 - R:api2.poloniex.com/104.18.216.39:443]

PR #487, opened 2020-01-02
Change update rate to 100ms for binance market data service

lepus:
I have changed the update speed for binance market data service from 1000ms (default) to 100ms


xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
```diff
@@ -174,7 +171,13 @@ public static String buildSubscriptionStreams(ProductSubscription subscription)
     }
 
     private static String buildSubscriptionStrings(List<CurrencyPair> currencyPairs, String subscriptionType) {
-        return subscriptionStrings(currencyPairs).map( s -> s + "@" + subscriptionType).collect(Collectors.joining("/"));
+        if ("depth".equals(subscriptionType)) {
```
badgerwithagun: Thanks for your contribution, @lepus!

This looks good, but I'd like to retain the existing behaviour and make this configurable instead.

This is straightforward.  XStream core provides a means to provide configuration key-value pairs when initialising the `Exchange`: you call `setExchangeSpecificParametersItem` on `ExchangeSpecification`. Let's use that to specify the binance orderbook update frequency in ms.

Perhaps call it `Binance_Orderbook_Update_Frequency_Ms`?

If there is a minimum or maximum Binance allows, probably good idea to validate for it when starting.

lepus: Thanks for your hints @badgerwithagun,

I've implemented the changes as you described with exchange specific parameters.

Please have a look on that.

xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -72,8 +71,10 @@
     private final AtomicBoolean fallenBack = new AtomicBoolean();
     private final AtomicReference<Runnable> fallbackOnApiCall = new AtomicReference<>(() -> {});
 
-    public BinanceStreamingMarketDataService(BinanceStreamingService service, BinanceMarketDataService marketDataService, Runnable onApiCall) {
+    public BinanceStreamingMarketDataService(BinanceStreamingService service, BinanceMarketDataService marketDataService, Runnable onApiCall,
```
badgerwithagun: For consistency, please follow the pattern used by `onApiCall`: extract the setting as an integer in `BinanceStreamingExchange.initServices` and then pass that integer here.

If you also perform the validation in `initServices` it will really simplify the code.  You can simply default the value there to 1000ms (use an `int`, not an `Integer`, so it can't be null) which means you don't need a lot of complex nested conditions when building the subscription string.

xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
```diff
@@ -165,16 +163,34 @@ private BinanceStreamingService createStreamingService(ProductSubscription subsc
...
                 buildSubscriptionStrings(subscription.getTrades(), "trade"))
                     .filter(s -> !s.isEmpty())
                     .collect(Collectors.joining("/"));
     }
 
-    private static String buildSubscriptionStrings(List<CurrencyPair> currencyPairs, String subscriptionType) {
-        return subscriptionStrings(currencyPairs).map( s -> s + "@" + subscriptionType).collect(Collectors.joining("/"));
+    private String buildSubscriptionStrings(List<CurrencyPair> currencyPairs, String subscriptionType) {
+        if ("depth".equals(subscriptionType)) {
+            Integer updateFrequency = (Integer) exchangeSpecification
```
badgerwithagun: If you simply had a `private int orderBookUpdateFrequency` extracted in `initServices`, which defaults to 1000ms, all this code would simplify down to a single line. You wouldn't need any `if` conditions at all.

Alternatively, if you want to default to whatever Binance provide by default, you could make `orderBookUpdateFrequency` a `String` which defaults to `""`.

xchange-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -119,9 +120,21 @@ public BinanceStreamingMarketDataService(BinanceStreamingService service, Binanc
             ));
     }
 
-    private static String channelFromCurrency(CurrencyPair currencyPair, String subscriptionType) {
+    private String channelFromCurrency(CurrencyPair currencyPair, String subscriptionType) {
         String currency = String.join("", currencyPair.toString().split("/")).toLowerCase();
-        return currency + "@" + subscriptionType;
+        String currencyChannel = currency + "@" + subscriptionType;
+
+        if ("depth".equals(subscriptionType)) {
```
badgerwithagun: It feels like this logic is now complex enough to warrant sharing between here and `BinanceStreamingExchange.buildSubscriptionStrings`.

badgerwithagun:
@lepus - this project is shortly to be merged with the main XChange project so we are clearing outstanding pull requests. Will you be able to implement the review changes?

lepus:
> @lepus - this project is shortly to be merged with the main XChange project so we are clearing outstanding pull requests. Will you be able to implement the review changes?

For sure. I think I will implement the review changes in the next few days. When do you plan to merge to main XChange?

PR #499, opened 2020-01-10
Logfile spam mitigation

mdvx:
I subscribe to a lot of tickers


mdvx:
BTW:  This warning could applies to a lot of exchanges, getting a good snapshot then updates is not always possible on all exchanges  (binance at least has a working solution)

PR #509, opened 2020-02-08
handle duplicate subscriptions warning

mdvx:
this is not a fatal error


mdvx:
https://docs.bitfinex.com/docs/ws-general

Error Codes
10300 : Subscription failed (generic)
10301 : Already subscribed
10302 : Unknown channel

10302 is probably ok for default processing

PR #511, opened 2020-02-11
[OkCoin] Implement v3 api for depth sub

e1ektr0:
Just depth


badgerwithagun:
Thanks for the contribution, @e1ektr0! Reviewing now.

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/WebSocketClientHandler.java
```diff
@@ -109,6 +109,7 @@ private void dealWithContinuation(ContinuationWebSocketFrame frame) {
 
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        cause.printStackTrace();
```
badgerwithagun: The right place to do this is by including it in the `LOG.error()` below:
```
LOG.error("WebSocket client encountered exception. Closing", cause);
```
... since `e.printStackTrace()` doesn't go through the "proper" log handler route, just direct to stdout, which will cause problems for people with special log handling requirements.

However, I'm pretty sure that if you look at the history of this code, you'll find that the above is exactly what it used to do, but we suppressed the actual stack trace since it ended up being too noisy. Unexpected errors on websockets turn out to happen an awful lot (the internet isn't reliable) it was spamming the logs too much.

Generally we try and catch exceptions in more case-specific ways.  Was this here to catch something in particular?

xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinStreamingExchange.java
```diff
@@ -60,3 +60,4 @@ public StreamingMarketDataService getStreamingMarketDataService() {
     @Override
     public void useCompressedMessages(boolean compressedMessages) { streamingService.useCompressedMessages(compressedMessages); }
 }
+
```
badgerwithagun: nit: try to avoid these sorts of whitespace/formatting changes in PRs. It makes the scope of a change harder to assess.

xchange-okcoin/src/test/java/info/bitrich/xchangestream/okcoin/OkCoinV3StreamingMarketDataServiceTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+public class OkCoinV3StreamingMarketDataServiceTest {
```
badgerwithagun: We avoid tests in the normal build talking directly to exchanges, since it means our builds fail if the exchange is down.  Tests like these are usually modelled as runnable "Example" classes (plenty of examples to follow).

One thing we should do more of is full-stack tests using utilities like Wiremock, if you're interested in giving that a go.  This code should have _some_ testing in the normal build.

xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinV3StreamingExchange.java
```diff
@@ -0,0 +1,62 @@
+package info.bitrich.xchangestream.okcoin;
```
badgerwithagun: Other than the different URL and the use of `OkCoinV3StreamingService` instead of `OkCoinStreamingService`, this class and `OkCoinV3StreamingMarketDataService` don't seem to be very different.  Could we instead introduce an exchange-specific parameter to switch between the two and avoid duplicating lots of code?

See `BitmexStreamingExchange` for an example. You could use an argument called something like `Use_OkCoin_V3_Api`.

xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinV3StreamingService.java
```diff
@@ -0,0 +1,155 @@
...
+                completable.onError(e);
+              }
+            });
+  }
+
+  @Override
+  protected String getChannelNameFromMessage(JsonNode message) throws IOException {
+    JsonNode table = message.get("table");
+    if (table != null) {
+      JsonNode data = message.get("data");
```
badgerwithagun: Worth being defensive against either `data` or `instrument_id` turning out to be absent?

xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinV3StreamingService.java
```diff
@@ -0,0 +1,155 @@
...
+      return result.replace("\"", "");
+    }
+    String event = message.get("event").asText();
+    return event;
+  }
+
+  @Override
+  public String getSubscribeMessage(String channelName, Object... args) throws IOException {
+
+    return String.format("{\"op\": \"subscribe\", \"args\": [\"%s\"]}", channelName);
```
badgerwithagun: nit: probably safer to create a `V3WebSocketMessage` class and use `ObjectMapper` to serialise it. (similarly to `OkCoinStreamingService`). Manually building JSON by concatenation can catch you out unexpectedly.  I think it's probably safe here but not good practice.

xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinV3StreamingService.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  @Override
+  public String getUnsubscribeMessage(String channelName) throws IOException {
+    return String.format("{\"op\": \"subscribe\", \"args\": [\"%s\"]}", channelName);
```
badgerwithagun: nit: probably safer to create a `V3WebSocketMessage` class and use `ObjectMapper` to serialise it. (similarly to `OkCoinStreamingService`). Manually building JSON by concatenation can catch you out unexpectedly.  I think it's probably safe here but not good practice.

xchange-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkCoinV3StreamingService.java
```diff
@@ -0,0 +1,155 @@
...
+    @Override
+    public void channelInactive(ChannelHandlerContext ctx) {
+      if (pingPongSubscription != null && !pingPongSubscription.isDisposed()) {
+        pingPongSubscription.dispose();
+      }
+      super.channelInactive(ctx);
+    }
+
+    @Override
+    public void channelRead0(ChannelHandlerContext ctx, Object msg) throws Exception {
```
badgerwithagun: Looks like you can share the one in `OkCoinStreamingService` (just move it to a new file) and add that extra check for `pong` at the end; it's probably safe for both versions.

badgerwithagun:
@e1ektr0 - this project will shortly be merging with the main XChange project so we need to close all outstanding PRs. Will you be able to follow up on the review comments above?

badgerwithagun:
This PR will be closed shortly unless progressed.

PR #512, opened 2020-02-15
[CoinbasePro] orderbook - this bug again

mdvx:
When system is loaded, bids and asks do not always get created, and since there is no synchronization we arrive in getOrderBook with bids.get(cuurencyPair) & asks.get(cuurencyPair) = null and message.getType().equals(L2UPDATE)


badgerwithagun:
Any answer on that, @mdvx ?

mdvx:
Looks like bitfinex dont always send the snapshot, I had not considered that but it makes total sense.

badgerwithagun:
OK, I'll merge this in

Issue #124, opened 2018-03-06
Emit all tickers


caleb-allen:
There are a handful of exchanges that allow you to subscribe to _all_ market tickers. The WS client is even getting all the data in the case of Poloniex2, and filters down the data to only return what has been subscribed to.

        return subscribedChannel
                .filter(s -> s.getPairId() == currencyPairId)
                .map(s -> adaptPoloniexTicker(s.toPoloniexTicker(currencyPair), currencyPair));

Is there a way to subscribe to all ticker events sent? I believe Binance and Bittrex support it as well, would be great to be able to handle with this library.

I'm also willing to implement that if I can be pointed in the right direction.

caleb-allen:
I believe this is the same issue seen in #98

caleb-allen:
Related to timmolter/XChange#2385

Issue #98, opened 2018-02-10
Connect all pairs on exchange


jurepetrovic:
Hey guys, 

Is there a way to subscribe to all pairs on an exchange? 
Or at least to put an array of currency pairs as a parameter when 
subscribing to StreamingExchange?

Thank you, 
Jure

Ameobea:
You can use `exchange.getExchangeSymbols()` to get a list of symbols supported for a given `StreamingExchange` instance.  Then, just use a for loop to subscribe to the data feeds for each of those symbols.

jurepetrovic:
Hello,

I know, I did that.
It's kind of funny, because it opens a new thread for each subscription and
then there are really a lot of them.
On some pairs there aren't many trades, so it's a little bit overthreaded
for my taste....

Regards,
Jure

2018-03-03 22:04 GMT+01:00 Casey Primozic <notifications@github.com>:

PR #520, opened 2020-02-27
Add getTickers for #124 and #98

TSavo:
Per tickets #98 and #124, it's been requested to allow for a single Observable that emits multiple (or all) currency pairs. This PR enhances the StreamingMarketData interface for all exchanges to support this with a new method: getTickers(CurrencyPair ...). Just pass it all the currencyPairs you are interested in, and on supported exchanges it will filter from a single stream, otherwise it will subscribe to all the tickers listed, and return a composite Observable that emits all the pairs requested.

### DIFF
--- a/xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
+++ b/xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
@@ -2,6 +2,7 @@ package info.bitrich.xchangestream.poloniex;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.MinMaxPriorityQueue;
+import com.google.common.collect.Sets;
 import info.bitrich.xchangestream.core.StreamingMarketDataService;
 import info.bitrich.xchangestream.poloniex.utils.MinMaxPriorityQueueUtils;
 import info.bitrich.xchangestream.service.wamp.WampStreamingService;
@@ -90,6 +91,12 @@ public class PoloniexStreamingMarketDataService implements StreamingMarketDataSe
 
     @Override
     public Observable<Ticker> getTicker(CurrencyPair currencyPair, Object... args) {
+        return getTickers(currencyPair);
+    }
+
+    @Override
+    public Observable<Ticker> getTickers(CurrencyPair... args) {
+        HashSet<CurrencyPair> pairs = Sets.newHashSet(args);
         return streamingService.subscribeChannel("ticker")
                 .map(pubSubData -> {
                     PoloniexMarketData marketData = new PoloniexMarketData();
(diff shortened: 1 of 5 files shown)


TSavo:
@badgerwithagun Requesting a review. :)

xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
```diff
@@ -90,6 +91,12 @@ public PoloniexStreamingMarketDataService(WampStreamingService streamingService)
 
     @Override
     public Observable<Ticker> getTicker(CurrencyPair currencyPair, Object... args) {
```
badgerwithagun: I've been thinking about this approach and how it combines with the principle that a good API should be a minimal one and not repeat itself unnecessarily.

Here's my thinking: the issue is not that the API doesn't provide what we want here.  As you have shown, you can achieve the same thing at the API level using:
```
Observable<Ticker> allTickers = Observable.fromIterable(List.of(tickers))
            .flatMap(pair -> exchange.getStreamingMarketDataService().getTicker(pair));
```
This is a one-liner, so the API is perfectly sufficient. The problem is actually that _the way the exchanges are supporting the existing API_ is inefficient.  This is endemic throughout xchange-stream.

If the implementation of the Poloniex `StreamingMarketDataService` only ever opened a single `1002` channel, `share()`ed it, and _re-used_ that existing `Observable` for any calls to `getTicker(...)`, with slightly different filters for each call but sharing the same underlying stream, then your approach and the above approach would be computationally equivalent, so there would be no performance penalty.

All of which is a long way to say: how about we just fix this obvious performance hole and avoid expanding the API?

xchange-stream-core/src/main/java/info/bitrich/xchangestream/core/StreamingMarketDataService.java
```diff
@@ -38,6 +40,24 @@
...
+     * Emits {@link info.bitrich.xchangestream.service.exception.NotConnectedException} When not connected to the WebSocket API.
+     *
+     * <p><strong>Warning:</strong> There are currently no guarantees
+     * that messages will not be skipped, or that any initial state
+     * message will be sent on connection.</p>
+     *
+     * @return {@link Observable} that emits {@link Ticker} when exchange sends the update.
+     */
+    default Observable<Ticker> getTickers(CurrencyPair... args) {
+        return Arrays.asList(args).stream()
```
badgerwithagun: The more RxJava way to express this is probably:

```
return Observable.fromIterable(Arrays.asList(args))
    .flatMap(pair -> getTicker(pair));
```

xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
TSavo: I follow! Let’s do that instead. I’ll Tag you again when it’s done.

PR #521, opened 2020-02-27
Change Poloniex to have a single subscription for Ticker. Fixes #124 and #98

### LINKED ISSUE #124: Emit all tickers

TSavo:
To better support #124 and #98, PoloniexStreamingMarketDataService has been changed to have a single subscription for Ticker which is then shared by all subscribers to the Ticker.

To get all the Tickers in a single observable, it is suggested the client create a list of all currencyPairs on the exchange, and use something like the following:

`Observable<Ticker> allTickers = Observable.fromIterable(currencyPairs).flatMap(pair -> streamingMarketDataService.getTicker(pair));`

### DIFF
--- a/xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
+++ b/xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
@@ -1,6 +1,8 @@
 package info.bitrich.xchangestream.poloniex;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.MinMaxPriorityQueue;
 import info.bitrich.xchangestream.core.StreamingMarketDataService;
 import info.bitrich.xchangestream.poloniex.utils.MinMaxPriorityQueueUtils;
@@ -22,10 +24,27 @@ import java.math.BigDecimal;
 import java.util.*;
 
 public class PoloniexStreamingMarketDataService implements StreamingMarketDataService {
+    public static final String TICKER_CHANNEL_NAME = "ticker";
     private final WampStreamingService streamingService;
+    private final Supplier<Observable<Ticker>> streamingTickers;
 
     public PoloniexStreamingMarketDataService(WampStreamingService streamingService) {
         this.streamingService = streamingService;
+        this.streamingTickers = Suppliers.memoize(() -> streamingService.subscribeChannel(TICKER_CHANNEL_NAME)
+                .map(pubSubData -> {
+                    PoloniexMarketData marketData = new PoloniexMarketData();
+                    marketData.setLast(new BigDecimal(pubSubData.arguments().get(1).asText()));
+                    marketData.setLowestAsk(new BigDecimal(pubSubData.arguments().get(2).asText()));
+                    marketData.setHighestBid(new BigDecimal(pubSubData.arguments().get(3).asText()));
+                    marketData.setPercentChange(new BigDecimal(pubSubData.arguments().get(4).asText()));
+                    marketData.setBaseVolume(new BigDecimal(pubSubData.arguments().get(5).asText()));
+                    marketData.setQuoteVolume(new BigDecimal(pubSubData.arguments().get(6).asText()));
+                    marketData.setHigh24hr(new BigDecimal(pubSubData.arguments().get(8).asText()));
+                    marketData.setLow24hr(new BigDecimal(pubSubData.arguments().get(9).asText()));
+
+                    PoloniexTicker ticker = new PoloniexTicker(marketData, PoloniexUtils.toCurrencyPair(pubSubData.arguments().get(0).asText()));
+                    return PoloniexAdapters.adaptPoloniexTicker(ticker, ticker.getCurrencyPair());
+                }).share());
     }
 
     private Map<CurrencyPair, MinMaxPriorityQueue<LimitOrder>> orderBookBids = new HashMap<>();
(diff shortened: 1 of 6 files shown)


TSavo:
@badgerwithagun Ditched the expanded API in favor of Poloniex having one shared observable. :)

xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
```diff
@@ -22,10 +22,27 @@
 import java.util.*;
 
 public class PoloniexStreamingMarketDataService implements StreamingMarketDataService {
+    public static final String TICKER_CHANNEL_NAME = "ticker";
     private final WampStreamingService streamingService;
+    private Observable<Ticker> streamingTickers;
 
     public PoloniexStreamingMarketDataService(WampStreamingService streamingService) {
         this.streamingService = streamingService;
+        this.streamingTickers = streamingService.subscribeChannel(TICKER_CHANNEL_NAME)
```
badgerwithagun: Rather than always subscribing to tickers even if they aren't needed, why not do this using `Suppliers.memoise()` from guava? Then the stream is initialised lazily if needed.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
```diff
@@ -23,13 +24,21 @@
...
-    private final Map<CurrencyPair, Integer> currencyPairMap;
+    private final Observable<Ticker> streamingTickers;
 
-    public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<CurrencyPair, Integer> currencyPairMap) {
+    public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<Integer, CurrencyPair> currencyIdMap) {
         this.service = service;
-        this.currencyPairMap = currencyPairMap;
+        final ObjectMapper mapper = StreamingObjectMapperHelper.getObjectMapper();
+
+        streamingTickers = service.subscribeChannel(TICKER_CHANNEL_ID)
```
badgerwithagun: Ditto

xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
TSavo: An outstanding idea. It is done.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
TSavo: It is done.

TSavo:
@badgerwithagun Great call on the memoizer. This has been rebased to develop and the memoization is now present and tested.

xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
```diff
@@ -22,10 +24,27 @@
 import java.util.*;
 
 public class PoloniexStreamingMarketDataService implements StreamingMarketDataService {
+    public static final String TICKER_CHANNEL_NAME = "ticker";
     private final WampStreamingService streamingService;
+    private Supplier<Observable<Ticker>> streamingTickers;
 
     public PoloniexStreamingMarketDataService(WampStreamingService streamingService) {
         this.streamingService = streamingService;
+        this.streamingTickers = Suppliers.memoize(streamingService.subscribeChannel(TICKER_CHANNEL_NAME)
```
badgerwithagun: Hm, this is fine but is making the (correct) assumption that `NettyStreamingService.subscribeChannel` doesn't actually do very much synchronously - the actual initialisation of the stream occurs when the `Observable` is subscribed, so we can afford to call `subscribeChannel` here without anything much actually happening as long as we memoise the call to `share()`.

This feels a bit brittle though, since `subscribeChannel` might change at some point to do a bit more in-line.  It's also easy to avoid - simply memoise the entire thing:

```
Suppliers.memoise(() -> streamingService.subscribeChannel()
   ...
   .share());
```

xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
```diff
@@ -22,10 +24,27 @@
 import java.util.*;
 
 public class PoloniexStreamingMarketDataService implements StreamingMarketDataService {
+    public static final String TICKER_CHANNEL_NAME = "ticker";
     private final WampStreamingService streamingService;
+    private Supplier<Observable<Ticker>> streamingTickers;
```
badgerwithagun: final

xchange-poloniex2/pom.xml
```diff
@@ -25,5 +25,10 @@
             <artifactId>service-netty</artifactId>
             <version>${project.parent.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>28.2-jre</version>
```
badgerwithagun: The version shou;ld be removed. It's already under `dependencyManagement` in the parent POM.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
```diff
@@ -23,13 +28,21 @@
...
-    private final Map<CurrencyPair, Integer> currencyPairMap;
+    private final Supplier<Observable<Ticker>> streamingTickers;
 
-    public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<CurrencyPair, Integer> currencyPairMap) {
+    public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<Integer, CurrencyPair> currencyIdMap) {
         this.service = service;
-        this.currencyPairMap = currencyPairMap;
+        final ObjectMapper mapper = StreamingObjectMapperHelper.getObjectMapper();
+
+        streamingTickers = Suppliers.memoize(service.subscribeChannel(TICKER_CHANNEL_ID)
```
badgerwithagun: Ditto

xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
TSavo: Done.

xchange-poloniex2/pom.xml
TSavo: Done.

xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
TSavo: Done.

xchange-poloniex/src/main/java/info/bitrich/xchangestream/poloniex/PoloniexStreamingMarketDataService.java
TSavo: Done.

TSavo:
@badgerwithagun All set. Memoized in it's entirety (not by method reference to share), final, and dependency versions removed.

PR #522, opened 2020-02-28
[core] subscribeChannel unnecessary onError

mdvx:
It seems to me that subscribeChannel can succeed even if the socket is not open and sendMessage already checks if it is open.
doOnDispose is unprotected from IO Exception?

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -328,8 +328,8 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
             if (channels.remove(channelId) != null) {
                 try {
                     sendMessage(getUnsubscribeMessage(channelId));
-                } catch (Exception e) {
-                    // ignore
+                } catch (IOException e) {
+                    LOG.warn("Failed to unsubscribe channel: {}",channelId);
                 }
             }
         }).share();
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -312,7 +312,7 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
         LOG.info("Subscribing to channel {}", channelId);
 
         return Observable.<T>create(e -> {
-            if (webSocketChannel == null || !webSocketChannel.isOpen()) {
+            if (webSocketChannel == null) { // || !webSocketChannel.isOpen()) {
                 e.onError(new NotConnectedException());
             }
             channels.computeIfAbsent(channelId, cid -> {
@@ -326,7 +326,11 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
             });
         }).doOnDispose(() -> {
             if (channels.remove(channelId) != null) {
-                sendMessage(getUnsubscribeMessage(channelId));
+                try {
+                    sendMessage(getUnsubscribeMessage(channelId));
+                } catch (Exception e) {
+                    // ignore
+                }
             }
         }).share();
     }


TSavo:
This looks good for merge, but I wouldn't mind if it was first rebased and fixedup into a single commit.

mdvx:
>  rebased and fixedup into a single commit.

I dont know how to do that, can you point me in right direction?

or maybe reject for now, and I will reapply after your other work

TSavo:
The basics: https://github.blog/2016-04-01-squash-your-commits/

The reason for this workflow: https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/

@mdvx

mdvx:
I am not sure, did that do it?

TSavo:
No there’s still 5 commits in this PR.

You want to:
git checkout branchName
git rebase -i HEAD~5
Squash all the commits
git —force push

mdvx:
can't get this to squash, but i like the concept and will use it in the future

TSavo:
It would be easiest at this point if upstream just used githubs rebase squash and merge.

It’s important to master this workflow. Imagine the following:

I want to test your fix in my branch, but because there’s now merges from development I’m having to pull only 3 of the 5 commits. If this were rebased first, the other two merge commits wouldn’t be there.

It’s time to do the release, and build the change log. I have five commits for one bug fix. Which commit message do I use? I have to read them all and understand what the change really is... vs one commit with one message, what the change is. If all PRs are squashed? You can automate the changelog off the commit messages.

Now imagine a bug occurs, and I think this may be responsible. I go to bisect the repo of this PR, and I have to figure out which commits to cherry pick, two of which are upstream merges that I want to keep. Compare that to just one commit, one bisect... my job is MUCH easier.

Now imagine I’m a downstream maintainer, and I want to merge in your change. I have to recognize that two commits are unrelated and should NOT be merged, but three do. Compare that to just one commit, properly rebased... no problem.

Of course the problem just gets worse when there’s more commits.

badgerwithagun:
I am quite happy squash-and-merging into master if there's a strong preference for it, but IMO the clarity of the merge history is a tooling issue, not a process one. A good git client can hide parents of a merge commit when viewing the history, giving the illusion of squash and merge having been used.

If someone thinks it is useful to show the sequence of commits in a PR (it can often help to understand the coding logic if they are particularly well-organised in their approach) I see no reason to prevent them doing this, nor any reason to hide that history after merge. If it was useful in review, it could also be useful later.

Personally I don't like my commits being squashed for the above reasons and also because it introduces complications when raising multiple PRs from the same branch - the branch has to be reset with prior commits skipped rather than cleanly fast-forwarded.

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -312,7 +312,7 @@ public void sendMessage(String message) {
         LOG.info("Subscribing to channel {}", channelId);
 
         return Observable.<T>create(e -> {
-            if (webSocketChannel == null || !webSocketChannel.isOpen()) {
+            if (webSocketChannel == null) { // || !webSocketChannel.isOpen()) {
```
badgerwithagun: Could you explain the logic here?

Also, if this check should be removed, it should be removed not commented out.

badgerwithagun:
> I want to test your fix in my branch, but because there’s now merges from development I’m having to pull only 3 of the 5 commits. If this were rebased first, the other two merge commits wouldn’t be there.
> 
> It’s time to do the release, and build the change log. I have five commits for one bug fix. Which commit message do I use? I have to read them all and understand what the change really is... vs one commit with one message, what the change is. If all PRs are squashed? You can automate the changelog off the commit messages.
> 
> Now imagine a bug occurs, and I think this may be responsible. I go to bisect the repo of this PR, and I have to figure out which commits to cherry pick, two of which are upstream merges that I want to keep. Compare that to just one commit, one bisect... my job is MUCH easier.
> 
> Now imagine I’m a downstream maintainer, and I want to merge in your change. I have to recognize that two commits are unrelated and should NOT be merged, but three do. Compare that to just one commit, properly rebased... no problem.

Hmm, I never have any problem with any of this. My git client handles most of it transparently. I just cherry-pick the merge commit (s) not the individual commits from the upstream branch.

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
mdvx: This is NettyStreamingService so I get the challenge.

| webSocketChannel == null 
Was there before and I do not challenge that logic

| || !webSocketChannel.isOpen()) {
This is unnecessary because sendMessage checks it too. 

Sorry I am just not that great at GIT (CVS was my tech horizon - I kid of course, i got on with subversion and I can check stuff into git - however, the complexity of these source control systems has developed to a point where you can either 'code it' or check-in 'correctly')

so lets just understand that 'dev-ops' is not 'dev' (and vice versa)

badgerwithagun: Don't worry about all the version control talk. That's not a question for now.

I don't think I agree that the second check isn't necessary. The emitter passed to `Observable.create()` is a callback which is called at the point a caller _subscribes_ to the observable. When an observable is first created, it is not subscribed - that happens when someone calls a terminating method on the observable chain - so there could be a lot of time between the `sendMessage` call and this callback getting invoked.

mdvx: let me take a second look, there is a good reason why i felt it was ok to remove it, mainly because it raised the error and 'failed' the subscription request.  It was my view, that you should not need a connection to make a subscription request (as it might be reconnecting while you are requesting - this case I came across).  But i know this is deep in the library, and could have other side affects.

badgerwithagun: Great. Happy to see it go if there's a good reason 👍

mdvx: after looking again, I think subscribe channel should look like this:
`public Observable<T> subscribeChannel(String channelName, Object... args) {
        final String channelId = getSubscriptionUniqueId(channelName, args);
        LOG.info("Subscribing to channel {}", channelId);

        return Observable.<T>create(e -> {
            channels.computeIfAbsent(channelId, cid -> {
                Subscription newSubscription = new Subscription(e, channelName, args);
                try {
                    // sendMessage will log an error if we are not connected, but it does not raise an exception
                    sendMessage(getSubscribeMessage(channelName, args));

                } catch (IOException throwable) { // if getSubscribeMessage throws this, it is because it needs to report
                    e.onError(throwable);    // a problem creating the message, should really be Exception (not IOException)
                }
                return newSubscription;
            });
        }).doOnDispose(() -> {
            if (channels.remove(channelId) != null) {
                try {
                    sendMessage(getUnsubscribeMessage(channelId));
                } catch (IOException e) {
                    LOG.error("Failed to unsubscribe channel: {}",channelId);
                }
            }
    }).share();
}`





As a separate issue, i would like to be able to disable the reconnect logic (so clients can be notified as soon as there is a problem with the socket), by maybe passing a negative retryDuration.


Not sure why this overrides the parameters passed
`
public NettyStreamingService(String apiUrl, int maxFramePayloadLength, Duration connectionTimeout, Duration retryDuration) {
//         this(apiUrl, maxFramePayloadLength, DEFAULT_CONNECTION_TIMEOUT, DEFAULT_RETRY_DURATION, DEFAULT_IDLE_TIMEOUT);

    this(apiUrl, maxFramePayloadLength, connectionTimeout, retryDuration, DEFAULT_IDLE_TIMEOUT)
}
`


` 
   private void scheduleReconnect() {
        if (retryDuration.toMillis() >= 0) {
            LOG.info("Scheduling reconnection");
            webSocketChannel.eventLoop().schedule(
                    () -> connect().subscribe(),
                    retryDuration.toMillis(),
                    TimeUnit.MILLISECONDS);
        } else {
            reconnFailEmitters.forEach(emitter -> emitter.onNext(new Exception("Reconnect disabled")));
        }
    }
`

I might have prefered to add another event, but i am not sure why reconnFailEmitters and connectionSuccessEmitters are lists instead of Subjects (legacy maybe)?

badgerwithagun:

badgerwithagun:

PR #523, opened 2020-02-28
Add Coveo Format plugin to maven and run it to format the code. Fixes #441

TSavo:
Trying to push the ball down the field here, this PR installs the Coveo format plugin for maven and contains the resulting formatted code.

Fixes #441, replaces #462 which requires rebasing.

I would prefer #524 to be merged first, and then we can easily redo the format on the renamed project.

PR #467, opened 2019-12-06
Rename projects to avoid clashing with XChange

makarid:
Fixes #464

PR #524, opened 2020-02-28
Rename all modules to xchange-stream per Issue #464

### LINKED ISSUE #464: All modules need renaming ahead of XChange merge
`service-xxx` should be renamed `xchange-stream-service-xxx`
`xchange-xxx` should be renamed `xchange-stream-xxx`

This will namespace the modules appropriately and avoid clash with existing XChange libraries.

TSavo:
In the interest of moving the ball down the field, I've opened this PR to replace #467 which needs rebasing.

Fixes #464, replaces #467.

I would prefer this to be merged before #523, and then we can easily redo the format on the renamed project.


TSavo:
This has been rebased against development.

PR #526, opened 2020-02-28
[BINANCE] Add specific param: use higher update frequency

lepus:
* Add exchange specific param to set update frequency (Binance_Orderbook_Use_Higher_Frequency). If set to true 100ms will be used as update frequency


lepus:
Hi @badgerwithagun, I've created a new clean pull request. Now I'm using only a boolean param, Because the allowed values are only 1000ms (default) and 100ms.
So tell me what else I should change.

Thanks a lot for your help.

PR #529, opened 2020-03-01
Bump to XChange 4.4.2-SNAPSHOT, fixing incompatibilities

badgerwithagun:
* Fixes Trade and UserTrade constructor usage
* Fixes change in name of LgoAdapter method
* Ignores BitmexTest which fails intermittently

### DIFF
--- a/xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
+++ b/xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
@@ -61,6 +61,10 @@ public class PoloniexStreamingMarketDataService implements StreamingMarketDataSe
     public Observable<Ticker> getTicker(CurrencyPair currencyPair, Object... args) {
         final ObjectMapper mapper = StreamingObjectMapperHelper.getObjectMapper();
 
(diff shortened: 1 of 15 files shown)


xchange-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingMarketDataService.java
```diff
@@ -61,6 +61,10 @@ public PoloniexStreamingMarketDataService(PoloniexStreamingService service, Map<
     public Observable<Ticker> getTicker(CurrencyPair currencyPair, Object... args) {
         final ObjectMapper mapper = StreamingObjectMapperHelper.getObjectMapper();
 
+        Observable<Ticker> tickerObservable = Observable.just(currencyPair)
```
TSavo: This feels like it shouldn't be here. It looks like it will result in a recursive call to itself, and the tickerObservable is never used.

badgerwithagun: Ooh, that definitely shouldn't be there. Crept in while I was checking some example code for one of your PRs....

badgerwithagun: Done, thanks for the spot

PR #530, opened 2020-03-02
Making all online tests @Ignore to allow build to work with no Internet connection. Fixes #357

### LINKED ISSUE #357: BitmexTest fails intermittently
This test connects to the actual exchange and asserts responses so will fail during periods of low activity or exchange downtime.

Example: https://travis-ci.org/bitrich-info/xchange-stream/builds/558927472

TSavo:
Sometimes tests that require an active connection to exchange cause the build to fail because of a problem on the exchange side. This PR changes all the online tests to @Ignore to avoid any possible build errors due to remote services being unavailable. Fixes #357 

To find these I literally unplugged my internet connection and ran all the tests, only @Ignore -ing the ones which failed.


badgerwithagun:
Or... Rather than using a profile, we could shift the matching tests to running in failsafe instead of surefire, and allow the build to pass if failsafe reports errors.

TSavo:
> Or... Rather than using a profile, we could shift the matching tests to running in failsafe instead of surefire, and allow the build to pass if failsafe reports errors.

I will look into that! :)

TSavo:
Closing this PR to do a failsafe integration instead.

PR #542, opened 2020-03-05
[core] add AutoReconnect control

mdvx:
Also one of the constructors was not passing parameters right

### DIFF
--- a/xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -88,7 +88,7 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
   // debugging
   private boolean acceptAllCertificates = false;
   private boolean enableLoggingHandler = false;
-  private boolean disableNettyReconnect = false;
+  private boolean autoReconnect = true;
   private LogLevel loggingHandlerLevel = LogLevel.DEBUG;
   private String socksProxyHost;
   private Integer socksProxyPort;
@@ -274,14 +274,14 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
   }
 
   private void scheduleReconnect() {
-    if (disableNettyReconnect) {
-        reconnFailEmitters.forEach(emitter -> emitter.onNext(new Exception("Netty reconnection disabled")));
+    if (autoReconnect) {
+      LOG.info("Scheduling reconnection");
+      webSocketChannel.eventLoop().schedule(
+              () -> connect().subscribe(),
+              retryDuration.toMillis(),
+              TimeUnit.MILLISECONDS);
     } else {
-        LOG.info("Scheduling reconnection");
-        webSocketChannel.eventLoop().schedule(
-                () -> connect().subscribe(),
-                retryDuration.toMillis(),
-                TimeUnit.MILLISECONDS);
+      reconnFailEmitters.forEach(emitter -> emitter.onNext(new Exception("AutoReconnect disabled")));
     }
   }
 
@@ -324,8 +324,13 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
   public String getSubscriptionUniqueId(String channelName, Object... args) {
     return channelName;
   }
+  /**
+   * Some exchanges rate limit messages sent to the socket (Kraken), by default this method does not rateLimit the
+   * messages sent out.
+   * Override this method to provide a rateLimiter, and call acquire on the rate limiter,
+   * to slow down out going messages.
+   */
   protected void sendMessageRateLimiterAcquire() {
-      // Some exchanges rate limit messages sent to the socket (Kraken)
       // no rate limiter by default
   }
   /**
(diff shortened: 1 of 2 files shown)


makarid:
The default will be to retry,is that right?

mdvx:
The default is to retry

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -529,4 +541,7 @@ public void setSocksProxyHost(String socksProxyHost) {
   public void setSocksProxyPort(Integer socksProxyPort) {
     this.socksProxyPort = socksProxyPort;
   }
+  public void setDisableNettyReconnect(boolean noRetry) {
```
TSavo: I'm not a fan of this naming of things as "setDisabled". It's a bit ambiguous as to what you're actually setting. The name also exposes the fact that it's Netty unnecessarily, where as this is an implementation detail specific to the fact that this is a NettyStreamingService. Imagine that it implements StreamingService, and the api comes from there... it wouldn't say Netty in the Interface. 

Can we instead make it, `setAutoReconnect(boolean autoReconnect)`, to make it perfectly clear that a true value will enable auto reconnect, where as a false value will make it not auto reconnect? Inside, we can translate that to Netty specific configuration, since this is a Netty implementation.

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -319,7 +324,10 @@ public Completable disconnect() {
   public String getSubscriptionUniqueId(String channelName, Object... args) {
     return channelName;
   }
-
+  protected void sendMessageRateLimiterAcquire() {
```
TSavo, line 324: I'm not clear on what this is doing.

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -385,12 +394,15 @@ public void sendMessage(String message) {
   }
 
   public void resubscribeChannels() {
-    for (Entry<String, Subscription> entry : channels.entrySet()) {
-      try {
-        Subscription subscription = entry.getValue();
-        sendMessage(getSubscribeMessage(subscription.channelName, subscription.args));
-      } catch (IOException e) {
-        LOG.error("Failed to reconnect channel: {}", entry.getKey());
+    if (channels.size() > 0) {
```
TSavo: This feels side-effecty. If you really do want to warn when this is called, it shouldn't not warn when the list is empty. If anything, I might expect the opposite case: a warning when resubscribe is called WITHOUT anything to resubscribe to.

Also, the if makes the whole method more indented for no reason: If the list is 0, the for will execute 0 times regardless, so there's no cost savings here, just more complexity of code.

mdvx: right now, a warning is displayed when the socket first connects, before any subscriptions are made, and this is normal behavior.  I dont think we need this warning at all, probably should only be a debug statement, if at all.  I would be quite happy to rid of that warning altogether.

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
mdvx: maybe needs a comment that this function should be overridden to apply a rate limiter.
It will make sense when I check in changes to Kraken.

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
mdvx: We can use AutoReconnect instead

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
TSavo: +1 for getting rid of it. :) And the if too.

mdvx:
I think I got all the comments @TSavo

mdvx:
any chance of getting some of the PRs approved?  I have a few more lined up behind, including things like Bittrex FOK/IOC/POSTONLY orders

badgerwithagun:
Will try. Routine slightly disrupted at the moment!

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -266,18 +267,22 @@ protected void initChannel(SocketChannel ch) {
...
-        .eventLoop()
-        .schedule(() -> connect().subscribe(), retryDuration.toMillis(), TimeUnit.MILLISECONDS);
+    if (autoReconnect) {
+      LOG.info("Scheduling reconnection");
+      webSocketChannel.eventLoop().schedule(
+              () -> connect().subscribe(),
+              retryDuration.toMillis(),
+              TimeUnit.MILLISECONDS);
+    } else {
+      reconnFailEmitters.forEach(emitter -> emitter.onNext(new Exception("AutoReconnect disabled")));
```
badgerwithagun: Hm, this is an interesting choice. I'm not sure about the idea of reporting a "reconnection failure" when no reconnection was attempted, or desired. Could you explain this choice?

PR #543, opened 2020-03-05
DisableNettyReconnect as ExchangeSpecification

mdvx:
parameter.
this depends on https://github.com/bitrich-info/xchange-stream/pull/542


xchange-stream-core/src/main/java/info/bitrich/xchangestream/core/StreamingExchange.java
```diff
@@ -15,6 +15,7 @@
   String ENABLE_LOGGING_HANDLER = "Enable_Logging_Handler";
   String SOCKS_PROXY_HOST = "SOCKS_Proxy_Host";
   String SOCKS_PROXY_PORT = "SOCKS_Proxy_Port";
+  String DISABLE_NETTY_RECONNECT = "No_Netty_Retry";
```
TSavo: I'd prefer to not expose the Netty aspect of this to higher API's and make it more descriptive. Perhaps, "Auto_Channel_Reconnect", where a true value enables reconnecting of channels, but a false disables it?

mdvx:
Moved to [core] add AutoReconnect control#542

PR #544, opened 2020-03-06
[core] Not all messages are in a channel

mdvx:
we see confirms an errors without a channel designation. 
@makarid I believe you were asking about this

### DIFF
--- a/xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -408,14 +408,16 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
 
   protected void handleMessage(T message) {
     String channel = getChannel(message);
-    if (channel != null)
+    if (channel != null && !channel.equals(""))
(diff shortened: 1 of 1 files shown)


xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -408,12 +408,14 @@ protected String getChannel(T message) {
 
   protected void handleMessage(T message) {
     String channel = getChannel(message);
-    handleChannelMessage(channel, message);
+    if (channel != null)
```
badgerwithagun: Hm, it doesn't look like valid behaviour for `getChannel()` to return `null`; `""` seems to be expected (which is what it does if there's an exception parsing the channel from the message). Perhaps instead change `getChannel()` to catch `null` and replace it with `""`?

mdvx: getChannel is overridden, so there is no knowing what it will return, "" or null or "a channel".  I agree probably dont want to send null or "" to handleChannelMessage.  I will add a check for "";

I added a log for handleError because I dont think we want that to fail silently.

PR #548, opened 2020-03-10
Resubscribe channels only when reconnecting

mladenmarkov:
Currently, the `resubscribeChannels()` is called on every successful connection.
This should be necessary only when reconnecting, i.e. in the `scheduleReconnect()` method.

### DIFF
--- a/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -224,18 +224,18 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
             }
             reconnFailEmitters.forEach(emitter -> emitter.onNext(t));
         })
-        .doOnComplete(() -> {
-            LOG.warn("Resubscribing channels");
-            resubscribeChannels();
-
-            connectionSuccessEmitters.forEach(emitter -> emitter.onNext(new Object()));
-        });
+        .doOnComplete(() -> connectionSuccessEmitters.forEach(emitter -> emitter.onNext(new Object())));
     }
 
     private void scheduleReconnect() {
         LOG.info("Scheduling reconnection");
         webSocketChannel.eventLoop().schedule(
-                () -> connect().subscribe(),
+                () -> connect().subscribe(
+                    () -> {
+                      LOG.warn("Resubscribing channels");
+                      resubscribeChannels();
+                    }
+                ),
                 retryDuration.toMillis(),
                 TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -224,18 +224,18 @@ protected void initChannel(SocketChannel ch) {
             }
             reconnFailEmitters.forEach(emitter -> emitter.onNext(t));
         })
-        .doOnComplete(() -> {
-            LOG.warn("Resubscribing channels");
-            resubscribeChannels();
-
-            connectionSuccessEmitters.forEach(emitter -> emitter.onNext(new Object()));
-        });
+        .doOnComplete(() -> connectionSuccessEmitters.forEach(emitter -> emitter.onNext(new Object())));
```
badgerwithagun, line 227: Could you check that this still works for exchanges such as Binance where channel subscriptions are made on connection?

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -224,18 +224,18 @@ protected void initChannel(SocketChannel ch) {
...
     }
 
     private void scheduleReconnect() {
         LOG.info("Scheduling reconnection");
         webSocketChannel.eventLoop().schedule(
-                () -> connect().subscribe(),
+                () -> connect().subscribe(
+                    () -> {
+                      LOG.warn("Resubscribing channels");
+                      resubscribeChannels();
```
badgerwithagun, line 236: This also needs to call the `connectionSuccessEmitters`.

badgerwithagun (changes requested):

service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
mdvx: can we move over to PublishSubject for reconnFailEmitters and connectionSuccessEmitters?  I know its a little out of scope, but would be nice.  Maybe emit the exchange too.

badgerwithagun: Funnily enough, I just suggested that on another review! Yes, that seems sensible - it means that subscriptions to those observables could persist between manual reconnects.

badgerwithagun:
This PR will be closed shortly unless there is further activity.

badgerwithagun:

badgerwithagun:

PR #550, opened 2020-03-10
[Kraken] Subscription failure detection

mladenmarkov:
Automatically detect subscription failure (for example, due to exceeded request rate) and send an `onError()` notification to subscriber.

Add a test to confirm that reliable subscriptions to all 135 Kraken pairs can be achieved, using a simple `retryWhen()` operator.

Depends on #549

### DIFF
--- a/xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
+++ b/xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
@@ -163,6 +163,19 @@ public class KrakenStreamingService extends JsonNettyStreamingService {
       }
       subscriptionRequestMap.put(reqID, channelName);
 
+      Observable.timer(2, TimeUnit.SECONDS)
+          .takeWhile(t -> isSocketOpen())
+          .subscribe(
+              t -> {
+                String chName = subscriptionRequestMap.get(reqID);
+                if (chName != null && chName.equals(channelName)) {
+                  LOG.error("Unable to subscribe channel {}", channelName);
+                  handleChannelError(
+                      channelName,
+                      new KrakenException("Unable to subscribe to channel " + channelName));
+                }
+              });
+
       KrakenSubscriptionMessage subscriptionMessage =
           new KrakenSubscriptionMessage(
               reqID,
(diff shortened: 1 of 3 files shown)


badgerwithagun:
Thanks for this @mladenmarkov!

Reviewing now.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -163,6 +163,19 @@ public String getSubscribeMessage(String channelName, Object... args) throws IOE
       }
       subscriptionRequestMap.put(reqID, channelName);
 
+      Observable.timer(2, TimeUnit.SECONDS)
+          .takeWhile(t -> isSocketOpen())
+          .subscribe(
+              t -> {
+                String chName = subscriptionRequestMap.get(reqID);
+                if (chName != null && chName.equals(channelName)) {
+                  LOG.error("Unable to subscribe channel {}", channelName);
```
badgerwithagun: We should be a little more explicit here, I think, just to be helpful, e.g. "Failed to receive subscription confirmation from Kraken within 2 seconds". And in the channel error.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -163,6 +163,19 @@ public String getSubscribeMessage(String channelName, Object... args) throws IOE
       }
       subscriptionRequestMap.put(reqID, channelName);
 
+      Observable.timer(2, TimeUnit.SECONDS)
```
badgerwithagun: I'm a little concerned about what happens if we either close down or remove the subscription between sending a subscription and this timer firing.  The latter probably isn't a big issue, but the former will result in some very odd errors.

Can we add something in here to skip the error if we've been shut down when the observable fires, or if the channel has been removed?

badgerwithagun (changes requested):

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
mladenmarkov: You are right. I improved the error message.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
mladenmarkov: If the connection is closed, then the `.takeWhile(t -> isSocketOpen())` would not pass and the error will not be generated.

If the consumer unsubscribed (i.e. disposed of the Disposable), then the `onError` handled will not be called at all.

I added a few tests to illustrate this behaviour. I think no odd errors are logged.
I may have missed a scenario, though. Let me know if you think of some.

badgerwithagun:

badgerwithagun:

PR #558, opened 2020-03-19
[coinbasepro] reconnectFailure & connectionSuccess NPE

mdvx:
return error for NPE instead of throwing one

### DIFF
--- a/xchange-stream-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
+++ b/xchange-stream-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
@@ -84,11 +84,17 @@ public class CoinbaseProStreamingExchange extends CoinbaseProExchange implements
 
   @Override
   public Observable<Throwable> reconnectFailure() {
+    if (streamingService == null)
+      return Observable.error(new NullPointerException("streamingService is null"));
+    
     return streamingService.subscribeReconnectFailure();
   }
 
   @Override
   public Observable<Object> connectionSuccess() {
+    if (streamingService == null)
+      return Observable.error(new NullPointerException("streamingService is null"));
+    
     return streamingService.subscribeConnectionSuccess();
   }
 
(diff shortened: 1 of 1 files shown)


xchange-stream-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
```diff
@@ -84,11 +84,17 @@ public Completable disconnect() {
 
   @Override
   public Observable<Throwable> reconnectFailure() {
+    if (streamingService == null)
+      return Observable.error(new NullPointerException("streamingService is null"));
```
badgerwithagun: How about an `IllegalStateException("Not connected")`? More descriptive.

badgerwithagun (changes requested):

xchange-stream-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
mdvx: works for me too

xchange-stream-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
```diff
@@ -84,11 +84,17 @@ public Completable disconnect() {
 
   @Override
   public Observable<Throwable> reconnectFailure() {
+    if (streamingService == null)
```
badgerwithagun, line 87: Sorry to mess you around, but I've just had another think about this. I think it's perfectly reasonable to expect these observables to fire on initial connection, not just on automated reconnections.  It seems wrong that we can't register to hear about connections/reconnections until _after_ we've connected. Could we just return publishers here and hook them up to the `StreamingService` before connecting?

mdvx:
replaced by https://github.com/bitrich-info/xchange-stream/pull/566

Issue #561, opened 2020-03-22
[HitBTC] OrderBook is missing support of timestamp


dp770:
Documentation: https://api.hitbtc.com/#order-book 
Stream URL: https://api.hitbtc.com/api/2/public/orderbook/ETHBTC
Example message:
`{
  "ask": [
    {
      "size": "0.088",
      "price": "0.046002"
    },
    {
      "size": "0.200",
      "price": "0.046800"
    }
  ],
  "bid": [
    {
      "size": "0.005",
      "price": "0.046001"
    },
    {
      "size": "0.200",
      "price": "0.046000"
    }
  ],
  "timestamp": "2018-11-19T05:00:28.193Z"
}`

PR #563, opened 2020-03-23
Issue #561: [HitBTC] OrderBook is missing support of timestamp

dp770:
Timestamp attribute is added to HitbtcOrderBookWithTimestamp (extension of HitbtcOrderBook)

### DIFF
--- a/xchange-stream-hitbtc/src/main/java/info/bitrich/xchangestream/hitbtc/dto/HitbtcOrderBookWithTimestamp.java
+++ b/xchange-stream-hitbtc/src/main/java/info/bitrich/xchangestream/hitbtc/dto/HitbtcOrderBookWithTimestamp.java
@@ -1,0 +1,40 @@
+package info.bitrich.xchangestream.hitbtc.dto;
+
+import org.knowm.xchange.hitbtc.v2.dto.HitbtcOrderBook;
+import org.knowm.xchange.hitbtc.v2.dto.HitbtcOrderLimit;
+
+import java.util.Date;
+
+public class HitbtcOrderBookWithTimestamp extends HitbtcOrderBook {
+
+    private final Date timestamp;
+
+    HitbtcOrderBookWithTimestamp(Date timestamp, HitbtcOrderLimit[] asks, HitbtcOrderLimit[] bids) {
+        super(asks, bids);
+        this.timestamp = timestamp;
+    }
+
+    public Date getTimestamp() {
+        return timestamp;
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder asks = new StringBuilder();
+        StringBuilder bids = new StringBuilder();
+
+        for (HitbtcOrderLimit ask : getAsks()) {
+            asks.append(ask).append(';');
+        }
+
+        for (HitbtcOrderLimit bid : getBids()) {
+            bids.append(bid).append(';');
+        }
+
+        return "HitbtcOrderBookWithTimestamp{" +
+                "asks=" + asks +
+                ", bids=" + bids +
+                ", timestamp=" + timestamp +
+                '}';
+    }
+}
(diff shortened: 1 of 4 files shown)


xchange-stream-hitbtc/src/main/java/info/bitrich/xchangestream/hitbtc/dto/HitbtcWebSocketOrderBook.java
```diff
@@ -35,16 +44,22 @@ private void createFromLevels(HitbtcWebSocketOrderBookTransaction orderbookTrans
     }
 
     sequence = orderbookTransaction.getParams().getSequence();
+    timestamp = parseISOTimestampInUTC(orderbookTransaction.getParams().getTimestamp());
```
badgerwithagun: This parsing is built into Java. You could just use `ZonedDateTime.parse(orderbookTransaction.getParams().getTimestamp()).toInstant().toEpochMilli()`

xchange-stream-hitbtc/src/main/java/info/bitrich/xchangestream/hitbtc/dto/HitbtcOrderBookWithTimestamp.java
```diff
@@ -0,0 +1,40 @@
+package info.bitrich.xchangestream.hitbtc.dto;
```
badgerwithagun, line 1: I can't see how this is being used.  I'd expect to see a corresponding change in https://github.com/bitrich-info/xchange-stream/blob/develop/xchange-stream-hitbtc/src/main/java/info/bitrich/xchangestream/hitbtc/HitbtcStreamingMarketDataService.java to feed the timestamp to clients?

badgerwithagun:
Thanksfor your contribution @dp770! Just a couple of thoughts/questions.

xchange-stream-hitbtc/src/main/java/info/bitrich/xchangestream/hitbtc/dto/HitbtcWebSocketOrderBook.java
dp770: I didn't know that. Thanks. Everyday we learn something new :-)
I'll change, test and submit correction.

dp770: Doesn't seem to work with UTC timestamp. Getting the following exception:
`Text '2020-05-03T18:26:28.714Z' could not be parsed: Unable to obtain ZonedDateTime from TemporalAccessor: {MicroOfSecond=714000, MilliOfSecond=714, NanoOfSecond=714000000, InstantSeconds=1588530388},ISO of type java.time.format.Parsed`
I've tried other built-in formatters which seems to support milliseconds, but having same issue. A'm I missing something?

dp770: Ignore the previous comment, I had to supply date format with time zone. Added it - works perfect. I'll commit soon.

badgerwithagun: I got interested in why you needed a formatter. Might possibly be [this JDK bug](https://bugs.openjdk.java.net/browse/JDK-8041360) because it works for me  on JDK11 without any special handing:

xchange-stream-hitbtc/src/main/java/info/bitrich/xchangestream/hitbtc/dto/HitbtcOrderBookWithTimestamp.java
dp770: Locally, I extended _HitbtcAdapters.adaptOrderBook_ method from the _org.knowm.xchange_ library to support timestamps. However, ideally, it should be updated at the source library, rather than "overridden" here. Having said that, I'm fine to commit it as well if it does make sense.

xchange-stream-hitbtc/src/main/java/info/bitrich/xchangestream/hitbtc/dto/HitbtcWebSocketOrderBook.java
dp770: No, I don't think it's related. Seems, that bug was fixed in 8u20. Actually, checking local history of changes, I guess I messed up with my code, which caused that. I had something like in this shell:
image

So, I changed the code to short and simple version as you suggested. Thanks.

badgerwithagun:

badgerwithagun:

PR #565, opened 2020-03-30
Prevent sending pong with a timestamp that is before Cex IO's own time.

mladenmarkov:
Sometimes the timestamp sent by Cex IO is up to a second ahead of the current time. Then our pong message carries a timestamp that is before Cex IO's own time. When this happens, Cex IO seems to ignore the pong and closes the connection 20-30 seconds later.

### DIFF
--- a/xchange-stream-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
+++ b/xchange-stream-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
@@ -12,6 +12,7 @@ import io.reactivex.CompletableOnSubscribe;
 import io.reactivex.Observable;
 import io.reactivex.subjects.PublishSubject;
 import java.io.IOException;
+import java.time.Instant;
 import org.knowm.xchange.currency.CurrencyPair;
 import org.knowm.xchange.dto.Order;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


xchange-stream-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
```diff
@@ -183,6 +184,13 @@ protected void handleMessage(JsonNode message) {
             }
             break;
           case PING:
+            // Sometimes the timestamp sent by Cex IO is up to a second ahead of current time. Then our pong sends a
+            // timestamp before Cex IO's own time.
```
badgerwithagun, line 188: Where's this timestamp? Part of the websocket frame outside the message? It's certainly not in `CexioWebSocketPongMessage`.

xchange-stream-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
```diff
@@ -183,6 +184,13 @@ protected void handleMessage(JsonNode message) {
...
+            // When this happens, Cex IO seems to ignore the pong and closes the connection 20-30 seconds later
+            long time = message.get("time").longValue();
+            while (Instant.now().isBefore(Instant.ofEpochMilli(time))) {
+              Thread.yield();
```
badgerwithagun, line 192: This is still basically busy-waiting an event loop thread. If we really need to wait it would be better simply to [schedule the response](https://netty.io/4.0/api/io/netty/util/concurrent/EventExecutorGroup.html#schedule-java.lang.Runnable-long-java.util.concurrent.TimeUnit-), but I'd like to understand why we can't just sync the clock that's generating the problematic timestamp.

mladenmarkov:
I'm sorry, I haven't fully tested this. I got it from a colleague, without fully getting into the details. Fact is, the disconnections stopped after this change. However, I need to investigate this further before continuing with this pull request. Sorry to have wasted your time.

badgerwithagun:
No worries @mladenmarkov.  Shall we keep this PR open?

badgerwithagun:
Closing this for mow @mladenmarkov.  Please ask if you need it re-opened or submit a new one.

PR #566, opened 2020-03-31
[Core]NettyStreamingService using PublishSubject for connect/fail handling

mdvx:
with an implementation for Coinbase-Pro

### DIFF
--- a/xchange-stream-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
+++ b/xchange-stream-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
@@ -92,7 +92,7 @@ public class CoinbaseProStreamingExchange extends CoinbaseProExchange implements
 
   @Override
   public Observable<Throwable> reconnectFailure() {
-    return streamingService.subscribeReconnectFailure();
+    return reconnectFailureSubject.share();
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


xchange-stream-coinbasepro/src/main/java/info/bitrich/xchangestream/coinbasepro/CoinbaseProStreamingExchange.java
```diff
@@ -24,6 +25,9 @@
   private CoinbaseProStreamingMarketDataService streamingMarketDataService;
   private CoinbaseProStreamingTradeService streamingTradeService;
 
+  private PublishSubject<Throwable> reconnectFailureSubject = PublishSubject.create();
```
badgerwithagun, line 28: I see the intent here of having a second pair of subjects, but the pattern with all other outputs from `Streaming...Service` implementations is that any streams are broken when `disconnect()` has been called. It's unusual that these two in particular would be able to persist between a `disconnect()` and `connect()`.

Personally, I think _all_ `Observable`s should stay "alive" between reconnects (which is what you've done here), but that fact that they don't _currently_ has probably made a lot of developers lazy such that they aren't calling `Disposable.dispose()` on their subscriptions before `disconnect`ing, so it could break some existing apps (with both old and new subscriptions firing).

I'm working on a generic `StreamingExchange` wrapper at the moment which you would use like this:
```
StreamingExchange generic = new GenericStreamAndPollExchangeWrapper(coinbaseStreamingExchange);
```
This is intended to do four things:
* Create a `StreamingExchange` from a non-streaming `Exchange` by polling in the background, so all apps can be built based on the streaming API even if it's not been built yet
* Combine streaming and polling so that even if the socket disconnects, you are guaranteed to get updates on subscribed `Observable`s via a poll update after reconnection
* Temporarily (until upstream fixes are applied) correct for differences in the implementations for different exchanges
* Ensure that any open `Observable` persists between reconnects.

Once this is done, we can retain the existing "contract" where all disposables are implicitly orphaned on disconnect, but where a persistent subscription is required, you can use the wrapper.

I'm happy to post up a work in progress if you're interested. In the meantime, it may be safer to keep the `PublishSubject` inside `NettyStreamingExchange` only. Thoughts?

badgerwithagun:

badgerwithagun:

